PR #17, opened 2017-12-03
views.py: Add last generated timestamp

andrewda:
This adds a UTC last generated timestamp to the view as well as how
long ago it was updated.

Closes https://github.com/coala/community/issues/7

## Screenshot

screen shot 2017-12-04 at 10 35 52 am

### DIFF
--- a/gci/views.py
+++ b/gci/views.py
@@ -1,4 +1,6 @@
 from django.http import HttpResponse
+from datetime import datetime
+from calendar import timegm
 
 from .students import get_students, get_linked_students
 
@@ -28,4 +30,15 @@ def index(request):
                  '<a href="https://github.com/{username}">{username}</a>'
                  .format(student_url=student_url, student_id=student_id,
                          username=username))
+
+    timestamp = datetime.now().strftime('%Y-%m-%d %H:%M:%S')
+    s.append('</ul><i id="time" class="timestamp" data-time="{unix}">'
+             'Last updated: {timestamp} '
+             '(<span id="ago" class="timeago"></span>)</i>'
+             .format(unix=timegm(datetime.utcnow().utctimetuple()),
+                     timestamp=timestamp))
+
+    s.append('<script src="static/timeago.js"></script>')
+    s.append('<script>loadTimeElements()</script>')
+
     return HttpResponse('\n'.join(s))
--- a/static/timeago.js
+++ b/static/timeago.js
@@ -1,0 +1,30 @@
+function generateTimeString(timestamp) {
+  var sec = ((new Date()).getTime() / 1000) - parseInt(timestamp);
+  var min = sec / 60;
+  var hour = min / 60;
+  var day = hour / 24;
+
+  var timeString = '';
+  if (day >= 1) {
+    timeString = Math.round(day) + ' days ago';
+  } else if (hour >= 1) {
+    timeString = Math.round(hour) + ' hours ago';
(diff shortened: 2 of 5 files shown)


gci/views.py
```diff
@@ -28,4 +30,26 @@ def index(request):
                  '<a href="https://github.com/{username}">{username}</a>'
                  .format(student_url=student_url, student_id=student_id,
                          username=username))
+    timestamp = datetime.now().strftime('%Y-%m-%d %H:%M:%S')
+    s.append('</ul><i id="time" data-time="{unix}">'
+             'Last updated: {timestamp} (<span id="ago"></span>)</i>'
+             .format(unix=timegm(datetime.utcnow().utctimetuple()),
+                     timestamp=timestamp))
+    s.append('<script>'
```
jayvdb: Move the script into a static 'asset' `.js` file , which is published to the deployed site, and use `<script src="...">`

gci/views.py
```diff
@@ -28,4 +30,26 @@ def index(request):
                  '<a href="https://github.com/{username}">{username}</a>'
                  .format(student_url=student_url, student_id=student_id,
                          username=username))
+    timestamp = datetime.now().strftime('%Y-%m-%d %H:%M:%S')
+    s.append('</ul><i id="time" data-time="{unix}">'
```
jayvdb: sprinkle with CSS classes so that other ppl can make it prettier without fiddling here.

gci/views.py
andrewda: Where should the asset file go? Do I also need to write something in the deploy.sh to copy over these asset files?

andrewda: Nevermind, Django made this much easier than I expected 👍

static/timeago.js
```diff
@@ -0,0 +1,20 @@
+var timestamp = document.getElementById('time').getAttribute('data-time');
+var ago = document.getElementById('ago');
+
+var sec = ((new Date()).getTime() / 1000) - parseInt(timestamp);
+var min = sec / 60;
+var hour = min / 60;
+var day = hour / 24;
+
+var timeString = '';
+if (day > 1) {
```
Monal5031: Shouldn't it be `>=` ?

andrewda: Well, day will only be exactly equal to 1 for about a second, so it's not really necessary. I could make these >= instead of > but the one second difference probably isn't worth it.

Monal5031: not 1 second, `30 mins` actually. `Math.round()` will round time `>=23.5` to `24`  😉

static/timeago.js
```diff
@@ -0,0 +1,20 @@
+var timestamp = document.getElementById('time').getAttribute('data-time');
```
jayvdb: put all of this into two function.

The first function should be a HTML unaware utility function which takes a `timestamp` and `ago` value, and returns the timestring.

The function second should be a function which is given two HTML element objects, one for the `time` and the other for `ago`, and this section function would call the first function to get the computed value and would put that value into the `ago` element.

Optionally add a third function which takes no arguments, fetches the desired elements, and calls the second function.

The second or third function should be invoked from inside the HTML ; worse case, use `body.onload` to invoke it.

static/timeago.js
Monal5031: Suppose the hour turns out to be something `>=23.5` then your logic will display `24 hours ago`

Issue #19, opened 2017-12-04
Django 2.0 breaks distill


jayvdb:
Upstream https://github.com/mgrp/django-distill/issues/7

PR #20, opened 2017-12-04
requirements.txt: Prevent use of Django 2.0

jayvdb:
It is incompatible with django-distill at the moment.

Fixes https://github.com/coala/community/issues/19


SanketDG:
Not that it really matters here, but requirements ideally should always be pinned to a specific version

Issue #25, opened 2017-12-05
Progress of solving & creating issues


LifetimeAlone:
Website may include a bar chart widget, which will show the dynamics of solving & creating issues. It will highlight how active the community has been during periods of time. 
The X-axis of the chart will be timeline (e.g. last 5 month), and Y-axis - amount of issues solved and created. To be visually intuitive, created & solved issues should be marked with different colors. Also, dates with positive tendency (rise in activity) may be marked green, dates with negative tendency - red.
Tools:
Bar chart creation: http://www.chartjs.org or https://d3js.org
Fetching data: IGitt has issue searching func. For each repo of the org, we can search issues by last update date and status (it must be closed).

jayvdb:
[GitMate.io](https://gitmate.io) thinks the contributor prediction failed because there are not enough contributors. Three contributors must have authored comments on at least three issues each.

nalinbhardwaj:
Probably a bit of an opinion, but I feel a line chart may be better for data representation?
Since we are trying to visualise change as well, and that would eliminate the need for a color for dates, making the graph more clearer.

Issue #26, opened 2017-12-06
Get the GitHub Org from the git remote info


jayvdb:
Split off from #1 

Using travis environment variables will work in Travis, but not locally, or on other build.  bad idea.

We should use `git remote -v` to fetch the remote name (there should be some IGitt functionality for this?), and then use https://github.com/retr0h/git-url-parse to extract the org name.

This functionality belongs in the main application, not in the `gci` app.

jayvdb:
[GitMate.io](https://gitmate.io) thinks the contributor prediction failed because there are not enough contributors. Three contributors must have authored comments on at least three issues each.

jayvdb:
This should be done in way that is re-usable ; e.g. `.travis.yml` currently sets `COMMIT_AUTHOR_EMAIL: "nobody@coala.io"` , but that domain name should also be derived. e.g. #29

Issue #1, opened 2017-12-01
Add org logo and a favicon


jayvdb:
As this repository should be usable for any org, ~~detect the name of the GitHub org  from the Travis environment variables~~, and fetch the org's logo.

And then convert the logo into a favicon.

jayvdb:
Updated title from favicon.ico to favicon , as other formats are good enough for most browsers

jayvdb:
> detect the name of the GitHub org  from the Travis environment variables

I've striken this, and moved it to https://github.com/coala/community/issues/26

PR #27, opened 2017-12-06
activity: Make activity page with graphs on issues

nalinbhardwaj:
Build scraper that scrapes [issues.json from gh-board](https://coala.github.io/gh-board/issues.json) and produces a JSON file just containing stats. This JSON file is used for displaying a static page with graphs, which can be toggled from a dropdown to display past year/month or week's data.
Example page:
screen shot 2017-12-07 at 2 59 04 am

<strike>I haven't worked with Django before, so please let me know if something is wrong, or there are better ways to do what I do.</strike> I've since removed all uses of Django, and am deploying it individually since I couldn't get it to work with the rest of the codebase neatly(getting some errors, although it worked fine individually)

The data is processed in python, and not directly in frontend to prevent browser overload(issues.json is > 27 MB)
Closes https://github.com/coala/community/issues/25

<strike> **TODO**: Once all changes are approved, make orderly commits with proper fragmentation, description etc. </strike>(done)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,6 @@ after_success:
   - python activity/scraper.py
   - python manage.py collectstatic --noinput
   - python manage.py distill-local public --force
+  - mkdir public/activity
+  - cp activity/index.html public/activity/
   - ./.ci/deploy.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
 before_script:
   - pip install coala-bears
   - npm i -g jshint
+  - bash orgname.sh
 
 script:
   - coala --non-interactive -V
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ script:
 
 after_success:
   - mkdir _site public
+  - python activity/scraper.py
   - python manage.py collectstatic --noinput
   - python manage.py distill-local public --force
   - ./.ci/deploy.sh
--- a/activity/index.html
+++ b/activity/index.html
@@ -1,0 +1,41 @@
+<!doctype html>
+<html>
+
+<head>
+    <title>Community Activity</title>
+    <style>
+    canvas {
+        -moz-user-select: none;
+        -webkit-user-select: none;
+        -ms-user-select: none;
+    }
+    </style>
+    <meta http-equiv="content-type" content="text/html; charset=utf-8" />
+</head>
+
+<body>
+    <h1>Community Activity</h1>
+    <br>
+    <br>
+    <div style="width:75%;">
+        <canvas id="canvas"></canvas>
+    </div>
+    <select id="chartType">
+        <option value="Month">Year</option>
+        <option value="Week">Month</option>
+        <option value="Day">Week</option>
+    </select>
+    <br>
+    <br>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.7.1/Chart.bundle.js"></script>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
+    <script src="/static/charts.js"></script>
+    <script>
+    $('#chartType').on('change', function(){
+        updateChart($('#chartType').val())
+    });
+    updateChart($('#chartType').val());
+    </script>
+</body>
+
+</html>
--- a/activity/scraper.py
+++ b/activity/scraper.py
@@ -1,0 +1,147 @@
+import requests
+import json
+import datetime
+import os
+import calendar
+from dateutil import parser, relativedelta
+
+
+class Scraper():
+    """
+    This is the class responsible for scraping provided issues into a
+    dictionary containing just statistical information of the data.
+    """
+
+    """
+    Count of months/weeks/days respectively to be scraped in past.
+    """
+    CONSTANTS = {
+        'month_count': 12,
+        'week_count': 4,
+        'day_count': 7,
+    }
+
+    def __init__(self, content, date):
+        """
+        Constructs a new ``Scraper``
+
+        :param content: Github API Parsed JSON issues
+        :param date: The date to scrape data till.
+        """
+        self.date = date
+        self.content = content
+
+        # Initialise data dicts
+        self.data = {
+            'year': {
+                'labels': [],
+                'closed': [0]*self.CONSTANTS['month_count'],
+                'opened': [0]*self.CONSTANTS['month_count'],
+            },
+            'month': {
+                'labels': [],
+                'closed': [0]*self.CONSTANTS['week_count'],
+                'opened': [0]*self.CONSTANTS['week_count'],
+            },
+            'week': {
+                'labels': [],
+                'closed': [0]*self.CONSTANTS['day_count'],
+                'opened': [0]*self.CONSTANTS['day_count'],
+            },
+        }
+
+        # Process labels for each option
+        for x in range(self.CONSTANTS['month_count']-1, -1, -1):
+            self.data['year']['labels'].append(calendar.month_name[(
+                self.date - relativedelta.relativedelta(months=x)).month])
+
+        for x in range(self.CONSTANTS['week_count']-1, -1, -1):
+            day = self.date - relativedelta.relativedelta(weeks=x)
+            strt = (day - datetime.timedelta(days=day.weekday()))
+            fin = (day + datetime.timedelta(days=6-day.weekday()))
+            self.data['month']['labels'].append(
+                calendar.month_abbr[strt.month] + ' ' + str(strt.day)
+                + ' - '
+                + calendar.month_abbr[fin.month] + ' ' + str(fin.day))
+
+        for x in range(self.CONSTANTS['day_count']-1, -1, -1):
+            day_idx = (self.date - datetime.timedelta(days=x)).weekday()
+            self.data['week']['labels'].append(calendar.day_name[day_idx])
+
+    def __diff_month(self, d):
+        """
+        :param d: Date as datetime, self.date >= Date.
+
+        :return: Difference in months(int) ignoring partially complete months.
+        """
+        return (self.date.year - d.year) * 12 + self.date.month - d.month
+
+    def __diff_week(self, d):
+        """
+        :param d: Date as datetime, self.date >= Date.
+
+        :return: Difference in weeks(int) ignoring partially complete weeks.
+        """
+        monday1 = (self.date - datetime.timedelta(days=self.date.weekday()))
+        monday2 = (d - datetime.timedelta(days=d.weekday()))
+        return (monday1 - monday2).days // 7
+
+    def __diff_days(self, d):
+        """
+        :param d: Date as datetime, self.date >= Date.
+
+        :return: Difference in days(int) ignoring partially complete days.
+        """
+        return (self.date-d).days
+
+    def get_data(self):
+        """
+        Get data
+
+        :return: Data in form of dict containing year, month, week data.
+        """
+        for issue in self.content:
+            issue = issue['issue']
+            # Parse date, while ignoring the timestamp.
+            dt = parser.parse(issue['createdAt'][:10])
+
+            mon = self.__diff_month(dt)
+            if mon < self.CONSTANTS['month_count']:
+                mon = self.CONSTANTS['month_count'] - mon
+                self.data['year']['opened'][mon] += 1
+                if issue['state'] == 'closed':
+                    self.data['year']['closed'][mon] += 1
+
+            wk = self.__diff_week(dt)
+            if wk < self.CONSTANTS['week_count']:
+                wk = self.CONSTANTS['week_count'] - wk
+                self.data['month']['opened'][wk] += 1
+                if issue['state'] == 'closed':
+                    self.data['month']['closed'][wk] += 1
+
+            dys = self.__diff_days(dt)
+            if dys < self.CONSTANTS['day_count']:
+                dys = self.CONSTANTS['day_count'] - dys
+                self.data['week']['opened'][dys] += 1
+                if issue['state'] == 'closed':
+                    self.data['week']['closed'][dys] += 1
+
+        return self.data
+
+
+if __name__ == '__main__':
+
+    org_name = open('org_name.txt').readline()
+
+    # URL to grab all issues from
+    issues_url = 'http://' + org_name + '.github.io/gh-board/issues.json'
+
+    content = requests.get(issues_url)
+    parsed_json = content.json()
+
+    real_data = Scraper(parsed_json['issues'], datetime.datetime.today())
+    real_data = real_data.get_data()
+
+    print(real_data)
+    with open('static' + os.sep + 'activity-data.json', 'w') as fp:
+        json.dump(real_data, fp)
--- a/orgname.sh
+++ b/orgname.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+regex="github.com/[a-z0-9A-Z]*"
+git config -l | grep -o "$regex" | while read -r line ; do
+    echo "${line:11}" | xargs echo -n > org_name.txt
+done
--- a/static/charts.js
+++ b/static/charts.js
@@ -1,0 +1,86 @@
+/* globals $, Chart */
+var curChart;
+
+function setChart(labels, openedData, closedData, type) {
+  var ctx = document.getElementById("canvas");
+
+  curChart = new Chart(ctx, {
+    type: 'line',
+    data: {
+      labels: labels,
+      datasets: [{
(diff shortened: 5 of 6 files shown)


activity/index.html
```diff
@@ -0,0 +1,131 @@
+<!doctype html>
+<html>
+
+<head>
+    <title>Coala Community Activity</title>
```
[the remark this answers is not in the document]
nalinbhardwaj: So, should I remove the word 'coala' or what?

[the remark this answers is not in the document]
blazeu: lower case c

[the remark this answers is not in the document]
blazeu: We don't allow "coala" here, the organization name must be dynamic.

By the way, coala is always written with lower case 'c'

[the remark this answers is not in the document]
Monal5031: We can keep this as `Organization Community Activity` for now, later after the model patch, this can be made dynamic.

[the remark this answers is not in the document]
blazeu: `# Ignore KeywordBear`

activity/scraper.py
```diff
@@ -0,0 +1,80 @@
...
+issuesurl = "https://coala.github.io/gh-board/issues.json"
+today = datetime.datetime.today()
+
+def diff_month(d1, d2):
+	return (d1.year - d2.year) * 12 + d1.month - d2.month
+
+def diff_week(d1, d2):
+	monday1 = (d1 - datetime.timedelta(days=d1.weekday()))
+	monday2 = (d2 - datetime.timedelta(days=d2.weekday()))
+	return (monday1 - monday2).days / 7
```
Techievena: Though I am not sure about this. I think this would be rather `// 7`. We need the floor value for the number of weeks right.

nalinbhardwaj: Monday1-monday2 is guaranteed to be a multiple of 7. Since, it’s a difference in Monday’s of different weeks...

Techievena: Then maybe `return ((d1-d2).days)//7` would be more simple.

nalinbhardwaj: The purpose behind doing it this way is to make sure weeks are counted started from Monday. So if the data is scraped on a Thursday, the current week is Monday-Thursday, and not previous Friday-Thursday, since it doesn’t make much sense mentally to have weeks not start from Monday, and also, it maintains consistency with other options(for example, year counts a month partially if it hasn’t ended yet)

nalinbhardwaj: I've left a comment in the code now to be clear(er) about this for future readers.

.travis.yml
```diff
@@ -17,4 +17,9 @@ after_success:
   - mkdir _site public
   - python manage.py collectstatic --noinput
   - python manage.py distill-local public --force
+  - mkdir public/activity
+  - python activity/scraper.py
```
jayvdb: Move all of this before the django stuff.
Then the json could be loaded by djanjo, but more importantly the static components should all be managed by djanjo's asset system. (e.g. there are django addons that do cache busting)

activity/index.html
```diff
@@ -0,0 +1,133 @@
...
+    </select>
+    <br>
+    <br>
+    <script>
+    var openedData = []
+    var closedData = []
+    var labels = [];
+    var myChart;
+
+    function updateChart() {
```
jayvdb: this should also be in static JS files, so they can be linted.

pass the document elements into the functions.

activity/scraper.py
```diff
@@ -0,0 +1,95 @@
+import requests
+import json
+import datetime
+from dateutil import parser, relativedelta
+
+# URL to grab all issues from
+issuesurl = "https://coala.github.io/gh-board/issues.json"  # Ignore KeywordBear
```
jayvdb: no, dont ignore the bears, they will eat you :P 
derive the `org_name` from your working environment.
See #1 and #26

activity/scraper.py
```diff
@@ -0,0 +1,95 @@
...
+	monday1 = (d1 - datetime.timedelta(days=d1.weekday()))
+	monday2 = (d2 - datetime.timedelta(days=d2.weekday()))
+	return (monday1 - monday2).days / 7
+
+# Calculate difference in days(ignores and hours difference(since both are dates))
+def diff_days(d1, d2):
+	return (d1-d2).days
+
+# Get labels for each option
+month_labels = []
```
jayvdb: you want the python std [`calendar`](https://docs.python.org/3/library/calendar.html) library, which is unfortunately not well known (and will be necessary if we add i18n).

Also move 80% of the remaining code into a class, even if it isnt well designed, because a class will force you to do some basic design, and then the indent levels will be correct , and it is easier to shuffle code around after that has been done.

activity/scraper.py
```diff
@@ -0,0 +1,95 @@
...
+			data["month"]["closed"][wk] += 1
+
+	dys = diff_days(today, dt)
+	if dys < day_count:
+		dys = day_count - dys
+		data["week"]["opened"][dys] += 1
+		if issue["state"] == "closed":
+			data["week"]["closed"][dys] += 1
+
+print(data)
```
jayvdb: hmm

activity/scraper.py
```diff
@@ -0,0 +1,95 @@
...
+with open('activity/data.json', 'w') as fp:
```
jayvdb: `os.sep`, for proper cross platform support.

activity/scraper.py
nalinbhardwaj: Printed for logging purposes currently, the rest of the codebase doesn't appear to use proper logging so I just printed it to console too :P

jayvdb:
unack https://github.com/coala/community/pull/27/commits/0f2b4a2be011fe7b9ba63122ecfe1cb3cc9ade68

.travis.yml
```diff
@@ -3,6 +3,7 @@ python: 3.6
 
 env:
   global:
+  - ORG_NAME: "coala"
```
jayvdb: Not allowed to put the word coala anywhere in this repo; it must work dynamically; when forked to a different org, it must derive the org.  I've also now created https://github.com/coala/community/issues/29 for the existing use of `coala` in this file.

activity/scraper.py
```diff
@@ -0,0 +1,134 @@
...
+    content = requests.get(issuesurl)
+    parsed_json = content.json()
+
+    real_data = Scraper(parsed_json["issues"], datetime.datetime.today())
+    real_data = real_data.get_data()
+
+    print(real_data)
+    with open('static' + os.sep + 'activity-data.json', 'w') as fp:
+        json.dump(real_data, fp)
+
```
jayvdb: EOL problem here.

static/charts.js
```diff
@@ -0,0 +1,89 @@
...
+                }]
+            }
+        }
+    });
+}
+
+function updateChart(type) {
+    if(curChart){ curChart.destroy(); }
+
+    $.getJSON({
```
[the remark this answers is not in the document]
jayvdb: I think you need to configure JSHintBear to accept JQuery's `$`.

static/charts.js
```diff
@@ -0,0 +1,89 @@
+var curChart;
+
+function setChart(labels, openedData, closedData, type) {
+    var ctx = document.getElementById("canvas");
+
+    curChart = new Chart(ctx, {
```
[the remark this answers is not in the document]
jayvdb: There should be a way to tell JSHintBear (or configure jshint) to ignore `Chart`.

static/charts.js
```diff
@@ -0,0 +1,89 @@
+var curChart;
+
+function setChart(labels, openedData, closedData, type) {
+    var ctx = document.getElementById("canvas");
+
+    curChart = new Chart(ctx, {  // Ignore JSHintBear
```
andrewda: You should be able to add `Chart` and `$` to the globals by adding this to the top of the file:

```
/* globals $, Chart */
```

nalinbhardwaj: Thanks, worked like a charm. :)

orgname.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+regex="github.com/[a-z0-9A-Z]*"
+git config -l | grep -o "$regex" | while read -r line ; do
+    echo "${line:11}" | xargs echo -n > org_name.txt
+done
```
Monal5031: newline here please.

nalinbhardwaj: fixed.

activity/index.html
```diff
@@ -0,0 +1,42 @@
...
+        -moz-user-select: none;
+        -webkit-user-select: none;
+        -ms-user-select: none;
+    }
+    </style>
+</head>
+
+<body>
+    <h1>Community Activity</h1>
+    <script>loadTimeElements()</script>
```
blazeu: Why not put it inside the same script tag below?

activity/index.html
```diff
@@ -0,0 +1,42 @@
+<!doctype html>
+<html>
+
+<head>
+    <title>Community Activity</title>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.7.1/Chart.bundle.js"></script>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
+    <script src="/static/timeago.js"></script>
+    <script src="/static/charts.js"></script>
```
blazeu: All of the `<script>` tag can be inside the body (below the content) so it doesn't block content.

`updateChart()` is called after the content anyway, so it doesn't matter.

activity/index.html
```diff
@@ -0,0 +1,42 @@
+<!doctype html>
+<html>
+
+<head>
+    <title>Community Activity</title>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.7.1/Chart.bundle.js"></script>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
```
blazeu: Is jquery 3 not compatible with Chart.js?

static/charts.js
```diff
@@ -0,0 +1,90 @@
...
+            else {
+                labels = data.week.labels;
+                openedData = data.week.opened;
+                closedData = data.week.closed;
+            }
+            setChart(labels, openedData, closedData, type);
+        },
+        error: function(error_data) {
+            console.log("error");
+            console.log(error_data);
```
blazeu: `console.error` is a bit accurate.

activity/index.html
nalinbhardwaj: Wasn't using it because it fails silently on `getJSON` errors(and I need logging), will change.

blazeu: I belive it's `.fail()` in jQuery 3

static/charts.js
```diff
@@ -0,0 +1,90 @@
...
+            }
+        }
+    });
+}
+
+function updateChart(type) {
+    if(curChart){ curChart.destroy(); }
+
+    $.getJSON({
+        url: "/static/activity-data.json",
```
blazeu: You can use the url directly `$.getJSON("/static/activity-data.json")`

activity/index.html
nalinbhardwaj: yep, changing.

static/charts.js
blazeu: ^ But it's optional.

orgname.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+regex="github.com/[a-z0-9A-Z]*"
+git config -l | grep -o "$regex" | while read -r line ; do
+    echo "${line:11}" | xargs echo -n > org_name.txt
```
jayvdb: I think we prefer two indents by default for shell.  Maybe see what ShellCheckBear says ;-)

activity/index.html
nalinbhardwaj: Actually, this is a remain of when I was serving the page using django. It's no longer possible to display the time stamp(since it's not generated via django)

blazeu:
This is just a comment.

When I check ChartJS, it does not need jQuery.

You need to stop using jQuery only for fetching things in the future :P
You can use the [native Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) and there's [500 bytes polyfill](https://github.com/developit/unfetch) for older browsers to support it.

For now I think it's ok since we probably need it for anything else, although I still prefer the native DOM API, it's getting better these days.

nalinbhardwaj:
@blazeu I'm currently also using `.on()`(although I could just replace that by an event listener). I think it's okay for it to use jQuery ATM. (I'll switch to a minified version for less loading).

orgname.sh
nalinbhardwaj: Seems not, [.ci/deploy.sh](https://github.com/coala/community/blob/master/.ci/deploy.sh) use 4 as well.

static/charts.js
```diff
@@ -0,0 +1,88 @@
...
+    });
+}
+
+function updateChart(type) {
+    if(curChart){ curChart.destroy(); }
+
+    $.getJSON("/static/activity-data.json",
+        function(data) {
+            var labels = data.year.labels;
+            var openedData = data.year.opened;
```
hemangsk: L65 and 66 are not needed right, its getting overwritten later inside if else anyway

nalinbhardwaj: Fixed.

orgname.sh
jayvdb: that was a horribly rushed script I mostly copied from other sources and fixed up; without code review; naughty me.

https://github.com/coala/coala-bears/blob/master/.ci/deps.sh
https://github.com/coala/coala/blob/master/.misc/deps.sh
(and others in those directories)

We dont run code linting on them yet.  I've created https://github.com/coala/coala/issues/4952 for that. (I'll publish a backlog of new tasks tomorrow)

jayvdb: (this script is so small it doesnt matter; you are right we dont have a convention yet, so you dont need to update this patch to follow a non-existent coding standard)

activity/scraper.py
```diff
@@ -0,0 +1,135 @@
+import requests
+import json
+import datetime
+import os
+import calendar
+from dateutil import parser, relativedelta
+
+class Scraper():
+    '''
```
jayvdb: `"""` for docstrings.

activity/scraper.py
```diff
@@ -0,0 +1,135 @@
...
+    This is the class responsible for scraping provided issues into a
+    dictionary containing just statistical information of the data.
+    '''
+
+    # Count of months/weeks/days respectively to be scraped in past.
+    month_count = 12
```
jayvdb: I am 99% sure these constants can be obtained from `calendar`.  If not, they should be CONSTANTS at the top of the module.

activity/scraper.py
```diff
@@ -0,0 +1,135 @@
...
+            "week" : {
+                "labels": [],
+                "closed": [0]*self.day_count,
+                "opened": [0]*self.day_count,
+            },
+        }
+
+        # Process labels for each option
+        for x in range(self.month_count-1, -1, -1):
+            self.data["year"]["labels"].append(calendar.month_name[(self.date - relativedelta.relativedelta(months=x)).month])
```
jayvdb: line length; please ensure this is being linted.

activity/scraper.py
```diff
@@ -0,0 +1,135 @@
...
+
+        :param content: Github API Parsed JSON issues
+        :param date: The date to scrape data till.
+        """
+        self.date = date
+        self.content = content
+
+        # Initialise data dicts
+        self.data = {
+            "year" : {
```
jayvdb: prefer single quotes in code.

activity/scraper.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        Calculate difference in days ignoring partially complete days.
+        
+        :param d: Date as datetime, self.date >= Date.
+        """
+        return (self.date-d).days
+
+    def get_data(self):
+        """
+        Get data
```
jayvdb: return type

activity/scraper.py
```diff
@@ -0,0 +1,135 @@
...
+                    self.data["week"]["closed"][dys] += 1
+
+        return self.data
+
+if __name__ == "__main__":
+
+    # URL to grab all issues from
+    org_name = open('org_name.txt').readline()
+
+    issuesurl = "http://" + org_name + ".github.io/gh-board/issues.json"
```
jayvdb: `issues_url`

activity/scraper.py
nalinbhardwaj: hmm, I chose these constants randomly myself. They're not representative of anything, just a number I think is good for display.(so no calendar use here)

jayvdb:
ack https://github.com/coala/community/pull/27/commits/6b69490819a3785502269c03e2c4a7ebc8572269

Issue #3, opened 2017-12-01
Build on forks with a static data file


jayvdb:
We can export a subset of the data as yaml/json, and store it in the repository, so that people can develop without access to a GCI API token

jayvdb:
Ideally https://github.com/coala/community/issues/2 is solved first.

jayvdb:
Some sample data can be found at https://github.com/i5o/gcileaderboard/tree/master/orgs/2016

jayvdb:
Load the data from there, so we do not keep a copy in our repository, as it is not our data.

jayvdb:
Now we have a data model, it is also possible to use https://github.com/gregmuellegger/django-autofixture

ShivamRai2003:
hey i want to work on this issue
can you help me what to do

PR #52, opened 2017-12-17
.coafile: Ignore coala in generated GCI files

jayvdb:
The GCI exported data contains 'coala',
so it needs to be excluded from the relevant checks.

Also ignore private/** and delete it regularly,
as a preventative measure for accidental leaks.

Related to https://github.com/coala/community/issues/3

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -7,6 +7,7 @@ mkdir private _site public
 if [[ -n "$GCI_TOKEN" ]]; then
   python manage.py fetch_gci_task_data private
   python manage.py cleanse_gci_task_data private _site
+  rm -rf private/
 else
   python manage.py fetch_old_gci_task_data _site || true
 fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ before_install:
 
 script:
   - ./.ci/build.sh
+  - rm -rf private/
   - coala --non-interactive -V
 
 after_success:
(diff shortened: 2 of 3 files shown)


.ci/build.sh
```diff
@@ -7,6 +7,7 @@ mkdir private _site public
 if [[ -n "$GCI_TOKEN" ]]; then
   python manage.py fetch_gci_task_data private
   python manage.py cleanse_gci_task_data private _site
+  rm -rf private/
```
yukiisbored: What's up with this?

.travis.yml
```diff
@@ -12,6 +12,7 @@ before_install:
 
 script:
   - ./.ci/build.sh
+  - rm -rf private/
```
yukiisbored: And this one as well...

.ci/build.sh
jayvdb: The private data collected on line 8 has been cleansed and pushed into `_site` , and is no longer needed.  It is what is causing the coala errors.  This is the first place to remove it.

.travis.yml
jayvdb: and in case somehow the directory is still existing here, it should be removed.  Extra careful.

yukiisbored:
ack https://github.com/coala/community/pull/52/commits/890cafa99b2fd320b99642d5ae4df5f1f8a7e292

Issue #9, opened 2017-12-01
Use logging


jayvdb:
Instead of print statements to the travis log, problems should be logged properly and a view generated with the logs.

abhinavralhan:
I would like to take this up @jayvdb . Any pointers how to get started ?

jayvdb:
I suggest using https://pypi.python.org/pypi/django-modelhandler to put the logging into a model , and then create a simple single page app that displays the data in the db.

seeeturtle:
thanks @jayvdb? I'll try it :)

PR #61, opened 2017-12-18
Replace every print to logging.

seeeturtle:
This will replace every print to logging.
It store log into /var/tmp/community.log.

Closes https://github.com/coala/community/issues/9

### DIFF
--- a/activity/scraper.py
+++ b/activity/scraper.py
@@ -3,7 +3,11 @@ import json
 import datetime
 import os
 import calendar
+
 from dateutil import parser, relativedelta
+from django.http import HttpResponse
+
+from community.git import get_owner
 
 
 class Scraper():
@@ -129,20 +133,20 @@ class Scraper():
         return self.data
 
 
-if __name__ == '__main__':
+def activity_json(request):
 
-    org_name = open('org_name.txt').readline()
+    org_name = get_owner()
 
     # URL to grab all issues from
     issues_url = 'http://' + org_name + '.github.io/gh-board/issues.json'
 
     content = requests.get(issues_url)
-    content.raise_for_status()
-    parsed_json = content.json()
+    try:
+        parsed_json = content.json()
+    except json.JSONDecodeError:
+        return HttpResponse('{}')
 
     real_data = Scraper(parsed_json['issues'], datetime.datetime.today())
     real_data = real_data.get_data()
 
-    print(real_data)
-    with open('static' + os.sep + 'activity-data.json', 'w') as fp:
-        json.dump(real_data, fp)
+    return HttpResponse(json.dumps(real_data))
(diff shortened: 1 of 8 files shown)


seeeturtle:
maybe we can add decorator for loading logging configuration. 
like:
```python
def load_logging_configuration(func):
  def wrapper():
    import logging.config
    logging.config.dictConfig(LOGGING_CONFIG)

@load_logging_configuration
def foo():
  logger = logging.getLogger(__name__ + foo.__name__)
  logger.info('bar')
```

jayvdb:
> maybe we can add decorator for loading logging configuration.

Ya, good idea.

activity/scraper.py
```diff
@@ -1,7 +1,6 @@
 import requests
 import json
 import datetime
-import os
```
jayvdb: unrelated to your issue ; please remove.  I've created https://github.com/coala/community/issues/62 for this.

gci/views.py
```diff
@@ -13,8 +14,34 @@
...
+            'level': 'NOTSET'
+        }
+    }
+}
+
 
 def index(request):
+    import logging.config
+
+    logging.config.dictConfig(LOGGING_CONFIG)
```
jayvdb: this should be in the main app `community`, not in the `gci` app.

gci/views.py
```diff
@@ -13,8 +14,34 @@
...
+        '': {
+            'handlers': ['communityHandler'],
+            'level': 'NOTSET'
+        }
+    }
+}
+
 
 def index(request):
+    import logging.config
```
jayvdb: imports belong at top of the file

log/view_log.py
```diff
@@ -0,0 +1,10 @@
+from django.http import HttpResponse
+
+def index(request):
+    logs = get_logs()
+    return HttpResponse('<br>'.join(logs))
```
jayvdb: https://deploy-preview-61--coala-community.netlify.com/log/ is currently empty

gci/views.py
```diff
@@ -13,8 +14,34 @@
...
+    'version': 1,
+    'formatters': {
+        'standard': {
+            'format': '%(asctime)s [%(levelname)s] %(name)s: %(message)s'
+        }
+    },
+    'handlers': {
+        'communityHandler': {
+            'class': 'logging.FileHandler',
+            'filename': '/var/tmp/community.log',
```
jayvdb: this file path should be stored in a separate constant variable.  Also this file path wont be valid on windows, so it will break any windows users.  you could simply place this file at `_site/log.txt` and it will be deployed to the live site.  not very pretty, but it is sufficient for this task.

log/view_log.py
seeeturtle: hmm... I'll check it.

gci/views.py
seeeturtle: how about `static/community.log`?

PR #63, opened 2017-12-18
Replace print into logging

seeeturtle:
This will replace print to standard logging.
It store logs in _site/community.log

Closes https://github.com/coala/community/issues/9

### DIFF
--- a/community/settings.py
+++ b/community/settings.py
@@ -12,6 +12,8 @@ https://docs.djangoproject.com/en/1.11/ref/settings/
 
 import os
 
+from .filters import NoDebugFilter
+
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
@@ -36,6 +38,36 @@ INSTALLED_APPS = [
     'django.contrib.staticfiles',
     'django_distill',
 ]
+
+LOGGING = {
+    'version': 1,
+    'disable_existing_loggers': False,
+    'filters': {
+        'noDebugFilter': {
+            '()': NoDebugFilter
+        }
+    },
+    'formatters': {
+        'standard': {
+            'format': '%(asctime)s [%(levelname)s] %(name)s: %(message)s'
+        }
+    },
+    'handlers': {
+        'communityHandler': {
+            'class': 'logging.FileHandler',
+            'filename': './_site/community.log',
+            'formatter': 'standard',
+            'filters': ['noDebugFilter'],
+            'level': 'INFO'
+        }
+    },
+    'loggers': {
+        '': {
+            'handlers': ['communityHandler'],
+            'level': 'INFO'
+        }
+    }
+}
 
 MIDDLEWARE = [
     'django.middleware.security.SecurityMiddleware',
--- a/gci/gitorg.py
+++ b/gci/gitorg.py
@@ -1,4 +1,5 @@
 import re
+import logging
 
 from .config import get_api_key
 
@@ -13,11 +14,12 @@ _orgs = {}
 class GitOrg():
 
     def __init__(self, name):
+        logger = logging.getLogger(__name__ + '.__init__')
         self.name = name
         self.IGH = GitHub(GitHubToken(get_api_key('GH')))
         self.IGL = GitLab(GitLabPrivateToken(get_api_key('GL')))
 
-        print('loading org %s' % name)
+        logger.info('loading org %s' % name)
         self.REPOS = dict()
 
         self.gh_repos = {repo.full_name.split('/')[-1]: repo for repo in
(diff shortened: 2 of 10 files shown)


manu-chroma:
It should be ``Replace print with`` (in your commit)

seeeturtle:
the log is empty in preview... what the matter?

log/view_log.py
```diff
@@ -0,0 +1,10 @@
+from django.http import HttpResponse
+
+def index(request):
+    logs = get_logs()
+    return HttpResponse('<br>'.join(logs))
+
+def get_logs():
+    with open('./_site/community.log') as log_file:
```
jayvdb: https://deploy-preview-63--coala-community.netlify.com/static/community.log is empty ;-(

community/decorators.py
```diff
@@ -0,0 +1,31 @@
+LOGGING_CONFIG = {
```
yukiisbored: Hey, Django already setup logging for us, Please don't create a new one, just use the one Django already setup.
https://docs.djangoproject.com/en/2.0/topics/logging/

community/decorators.py
```diff
@@ -0,0 +1,31 @@
...
+    'loggers': {
+        '': {
+            'handlers': ['communityHandler'],
+            'level': 'NOTSET'
+        }
+    }
+}
+
+
+def load_logging_configuration(func):
```
yukiisbored: Ref https://github.com/coala/community/pull/63/files#r157732719

community/decorators.py
yukiisbored: If you need to add more handlers, see https://docs.djangoproject.com/en/2.0/topics/logging/#configuring-logging,
Do it inside settings.py

jayvdb:
Please change the log displayed to not include debug.
And squash your commits.

community/settings.py
```diff
@@ -37,6 +37,30 @@
...
+        'standard': {
+            'format': '%(asctime)s [%(levelname)s] %(name)s: %(message)s'
+        }
+    },
+    'handlers': {
+        'communityHandler': {
+            'class': 'logging.FileHandler',
+            'filename': './_site/community.log',
+            'formatter': 'standard',
+            'level': 'NOTSET'
```
jayvdb: Set the level to INFO please.

gci/client.py
```diff
@@ -18,14 +18,12 @@
 
 client = gciclient.GCIAPIClient(
       auth_token='xxxxxxxxxxxxxx',
-      url_prefix='https://codein.withgoogle.com',
-      debug=False)
+      url_prefix='https://codein.withgoogle.com')
```
jayvdb: this is not the problem.  this is a docstring. please revert this bit.

gci/client.py
```diff
@@ -49,21 +47,13 @@ class GCIAPIClient(object):
   """
 
   def __init__(self, auth_token=None,
-               url_prefix='https://codein.withgoogle.com/',
-               debug=False):
```
jayvdb: revert this please.  We still need debug support, just not enabled by default

gci/client.py
```diff
@@ -49,21 +47,13 @@ class GCIAPIClient(object):
...
+      url_prefix='https://codein.withgoogle.com/'):
     self.url_prefix = urlparse.urljoin(url_prefix, 'api/program/current/')
     self.headers = {
         'Authorization': 'Bearer %s' % auth_token,
         'Content-Type': 'application/json',
     }
 
-    if debug:
```
jayvdb: please revert this.  debug is not enabled.  And is not enabled at https://github.com/coala/community/blob/master/gci/students.py#L30

community/settings.py
```diff
@@ -37,6 +37,30 @@
...
+    'disable_existing_loggers': False,
+    'formatters': {
+        'standard': {
+            'format': '%(asctime)s [%(levelname)s] %(name)s: %(message)s'
+        }
+    },
+    'handlers': {
+        'communityHandler': {
+            'class': 'logging.FileHandler',
+            'filename': './_site/community.log',
```
yukiisbored: Hey, put this inside the static folder.
https://docs.djangoproject.com/en/2.0/howto/static-files/

log/view_log.py
yukiisbored: This is not how you serve a static file,
read https://docs.djangoproject.com/en/2.0/howto/static-files/

community/settings.py
jayvdb: this isnt a pre-existing static file.  It has to go in `_site`

gci/students.py
```diff
@@ -19,6 +20,7 @@
     'deadline',
 )
 
+
```
jayvdb: undo this unrelated change

gci/views.py
```diff
@@ -1,12 +1,14 @@
...
 from datetime import datetime
 from calendar import timegm
+import logging
 import requests
 
 from .students import get_linked_students
 from .gitorg import get_logo
 from .task import get_tasks
 
+
```
jayvdb: undo this unrelated change

log/view_log.py
```diff
@@ -0,0 +1,11 @@
...
+        for line in log_file:
+            if 'DEBUG' not in line:
```
jayvdb: no this is not how you filter the log.  `logging` and django have official mechanisms for filtering the log.

We need things like access tokens to be never stored anywhere, at all.  Leaving them lying around in files is begging for leaks to occur again due to mistakes.

community/settings.py
seeeturtle: why? we have some log with `ERROR` level. like here: https://github.com/seeeturtle/community/blob/af7306e3e886d38d230a50efafa790f7a26221cc/gci/config.py#L33

log/view_log.py
seeeturtle: ok. I'll follow the official mechanisms

community/settings.py
jayvdb: This specifies the *minimum* level, and all levels above that are included.

gci/config.py
```diff
@@ -1,5 +1,7 @@
 import ruamel.yaml
 import os
+import logging
+
```
jayvdb: this extra blank line isnt desirable.

community/settings.py
seeeturtle: ah, I see. I didn't know well, sorry.

community/settings.py
```diff
@@ -11,6 +11,7 @@
 """
 
 import os
+from .filters import NoDebugFilter
```
jayvdb: add a blank line after system libraries, before local imports.

gci/config.py
```diff
@@ -27,6 +29,7 @@ def get_api_key(name):
             api_key = api_key_file.readline().strip()
             return api_key
     except IOError:
+        logger.error('Please put your %s API key at %s.' % (name, filename))
```
jayvdb: as a general rule, do not log before an exception.

An exception is telling the caller there was a problem, and the caller then has to decide what to do with the exception.
The caller may already know about this type of exception, and decide to ignore it because they do not consider it to be an error.

community/settings.py
seeeturtle: umm... you mean add two blanks between `import os` and `from .filters ...` or a blank

gci/config.py
```diff
@@ -15,6 +16,7 @@ class TokenMissing(KeyError):
 
 
 def get_api_key(name):
+    logger = logging.getLogger(__name__ + 'get_api_key')
```
jayvdb: logging names are dot separated.  so this should be `__name__ + '.get_api_key'` .  The same applies everywhere in your patch.

gci/config.py
seeeturtle: ok

gci/config.py
seeeturtle: ok

gci/gitorg.py
```diff
@@ -13,11 +14,12 @@
 class GitOrg():
 
     def __init__(self, name):
+        logger = logging.getLogger(__name__ + '__init__')
```
jayvdb: you need to add dots to *all* of the loggers.
And please rebase before asking for reviews.

seeeturtle: sorry, I missed it.

seeeturtle: done @jayvdb

jayvdb:
ack https://github.com/coala/community/pull/63/commits/fec438a9599de49eca25260866a9dfead9bcf152

Issue #58, opened 2017-12-18
Add GSOC data model


jayvdb:
GCI has [an API](https://developers.google.com/open-source/gci/api).

There is also a 'program' API for GSOC information at https://summerofcode.withgoogle.com/api/program/ ; the same data is available at https://summerofcode.withgoogle.com/api/program/2017/

That only provides information about the completed 2017 program, which is worth including, as preparation for the 2018 program.

For the 2018 program, there is [an ical](http://www.google.com/calendar/ical/gsummerofcode%40gmail.com/public/basic.ics) which can also be imported into the gsoc data model so we have at least basic information about the next GSOC program able to be used in our community website.

jayvdb:
The 2017 data is now disappeared ;-(

https://summerofcode.withgoogle.com/api/program/2018/ now has real data.

PR #64, opened 2017-12-18
Add GSOC data model

sks444:
Closes https://github.com/coala/community/issues/58


sks444:
> Can you try to populate your model with the 2017 JSON?

Yes, I could. But, I am not able to run any command because of that IGitt problem.

jayvdb:
You dont need IGitt to load JSON :P (comment out the gci app while testing)

sks444:
Where the data will go after I will load the json? We need a database right? 
>Also I'd prefer the dates are grouped into time periods, using the start&end dates with common names. This is a calendar / timeline, and the model should reflect that.

I didn't get this part, can you give a example?

gsoc/models.py
```diff
@@ -0,0 +1,222 @@
...
+    second_review_status_visible = models.BooleanField()
+    project_decisions_can_be_finalized = models.BooleanField()
+    student_home_address_is_editable = models.BooleanField()
+    result = models.ForeignKey(Result, on_delete=models.CASCADE)
+
+    def __str__(self):
+        return str(self.result)
+
+
+class Copy(models.Model):
```
hemangsk: Why is this model called Copy? 
PS: ignore me if this name has been discussed on gitter

hemangsk: Makes sense to call it something like Statistics?

gsoc/models.py
```diff
@@ -0,0 +1,222 @@
...
+    project_decisions_finalization = models.BooleanField()
+    first_work_period = models.BooleanField()
+    first_evaluations = models.BooleanField()
+    second_work_period = models.BooleanField()
+    second_evaluations = models.BooleanField()
+    third_work_period = models.BooleanField()
+    final_week = models.BooleanField()
+    final_evaluations_mentor = models.BooleanField()
+    post_program = models.BooleanField()
+    org_signup_open_starts = models.DateField()
```
jayvdb: I'd prefer the dates are grouped into time periods.

This is a timeline of periods.

See https://developers.google.com/open-source/gsoc/timeline

"org_signup" should have a start and end date that are grouped together

"org_approval" should have a start and end date that are grouped together

etc

gsoc/models.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+class Gsoc(models.Model):
+    count = models.IntegerField(primary_key=True)
+
+    def __str__(self):
+        return str(self.count)
+
+
+class Result(models.Model):
```
jayvdb: This model should be call Program.

gsoc/models.py
jayvdb: This should be ProgramStatistics

gsoc/models.py
```diff
@@ -0,0 +1,222 @@
...
+    def __str__(self):
+        return self.name
+
+
+class Project(models.Model):
+    name = models.CharField(max_length=100, primary_key=True)
+    summary = models.TextField(max_length=300, default='')
+    org = models.ForeignKey(Organization, on_delete=models.CASCADE)
+    student = models.OneToOneField('Student')
+    mentor = models.OneToOneField('Mentor')
```
jayvdb: many mentors per project

gsoc/models.py
sks444: @jayvdb,  I am not able to find a solution for how I group two dates together in Django model? Please help.

gsoc/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
```
jayvdb: I think this file may not be needed

gsoc/models.py
jayvdb: @hemangsk @nkprince007 @mamat-rahmat , any ideas on how to merge these start & end dates.

gsoc/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
```
jayvdb: I think we shouldnt have this file if we dont have any tests written

gsoc/views.py
```diff
@@ -0,0 +1,3 @@
+from django.shortcuts import render
```
jayvdb: likewise, can we avoid creating this file here, as it will be created in the other PR.

gsoc/models.py
```diff
@@ -0,0 +1,222 @@
...
+    def __str__(self):
+        return self.name
+
+
+class Project(models.Model):
+    name = models.CharField(max_length=100, primary_key=True)
+    summary = models.TextField(max_length=300, default='')
+    org = models.ForeignKey(Organization, on_delete=models.CASCADE)
+    student = models.OneToOneField('Student')
+    mentor = models.ForeignKey('Mentor', on_delete=models.CASCADE)
```
jayvdb: `Mentor` and `Student` models can appear above the `Project` model, so that they are hard references to the model instead of string references to the model.

gsoc/models.py
jayvdb: I think we need to use an existing django app or library to provide the data fields, and thus also hold the data.

https://djangopackages.org/grids/g/calendar/ shows the main options.

django-scheduler looks like it has the most features built in, while django-eventtools looks like the most raw & re-usable with fewer features.

Assuming django-eventtools is our best option, we would need a custom `GsocEvent(BaseEvent)` and `GsocOccurrence(BaseOccurrence)`, and the `Program` model would then need to have either
- an array of `GsocOccurrence` , and all of these start/end are pushed into it as events with the event name derived from the prefix (e.g. `org_signup_open` -> title="Org signup open")
- fields for each event, like `org_signup_open = GsocOccurrence()`

the later feels like the better option, at least initially.

gsoc/models.py
```diff
@@ -0,0 +1,222 @@
...
+    name = models.CharField(max_length=100)
+
+    def __str__(self):
+        return self.name
+
+
+# for 2018 events
+
+
+class UpcomingEvent(models.Model):
```
jayvdb: these should be GsocEvent/ GsocOccurrence - they are the same events that are found in the `Program` json

gsoc/models.py
jayvdb: You should now change this to `Mentor` instead of `'Mentor'`

gsoc/models.py
```diff
@@ -0,0 +1,205 @@
...
+    project_decisions_finalization = models.BooleanField()
+    first_work_period = models.BooleanField()
+    first_evaluations = models.BooleanField()
+    second_work_period = models.BooleanField()
+    second_evaluations = models.BooleanField()
+    third_work_period = models.BooleanField()
+    final_week = models.BooleanField()
+    final_evaluations_mentor = models.BooleanField()
+    post_program = models.BooleanField()
+    org_signup_open = models.OneToOneField('GsocEvent', related_name='org_signup_open')
```
jayvdb: Like "Mentor", put the GsocEvent and GsocOccurence above `Program`, so you can refer to the object instead of the name.

gsoc/models.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def __str__(self):
+        return self.event
+
+
+class Program(models.Model):
+    year = models.IntegerField(primary_key=True)
+    state = models.CharField(max_length=100)
+    is_current = models.BooleanField()
+    org_signup_open = models.BooleanField()
```
jayvdb: you have a second `org_signup_open` below ;-)

Many of these bools are like that; I guess they are derived ; i.e. this bool is enabled if the current date is in the date range for the event below

gsoc/models.py
```diff
@@ -0,0 +1,204 @@
...
+    results_announced = models.DateField()
+    tax_forms_deadline = models.DateField()
+    first_payments = models.DateField()
+    second_payments = models.DateField()
+    final_payments = models.DateField()
+    first_evaluations_finalized = models.DateField()
+    second_evaluations_finalized = models.DateField()
+    final_evaluations_finalized = models.DateField()
+    projects_accepted_count = models.IntegerField()
+    org_accepted_count = models.IntegerField()
```
jayvdb: move these two int fields to the top.

gsoc/models.py
```diff
@@ -0,0 +1,204 @@
...
+    slot_allocation_finalized = models.DateField()
+    second_enrollment_form_upload_deadline = models.DateField()
+    gsoc = models.ForeignKey(Gsoc, on_delete=models.CASCADE)
+
+    def __str__(self):
+        return str(self.year)
+
+
+class Feature(models.Model):
```
jayvdb: The `Feature` set are defined as part of the `Program` ; 1-1 relationship by the look of things.  Needs a foriegn key here.

As such, it should be called `Features` , and could even be inlined into the `Program`.

The alternative is to define a `Feature` which is a name and a bool , and then `Program` would have an attribute `features`.

gsoc/models.py
```diff
@@ -0,0 +1,204 @@
...
+    homepage_start_button = models.CharField(max_length=100)
+    homepage_intro_paragraph = models.TextField()
+    number_of_mentors = models.IntegerField()
+    number_of_student_and_mentor_countries = models.IntegerField()
+    number_of_years = models.IntegerField()
+    homepage_orgs_paragraph = models.TextField()
+    homepage_students_paragraph = models.TextField()
+    number_of_organizations = models.IntegerField()
+    number_of_student_countries = models.IntegerField()
+    program = models.ForeignKey(Program, on_delete=models.CASCADE)
```
jayvdb: foriegn key at top

gsoc/models.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def __str__(self):
+        return self.homepage_start_button
```
jayvdb: there is no good attribute to use for the `str()`.  can we use the program year ; it is the foreign key ??

gsoc/models.py
sks444: > you have a second org_signup_open below ;-)

Then, what name should I give the second one?

gsoc/models.py
```diff
@@ -0,0 +1,204 @@
...
+    org_payment_details_visible = models.BooleanField()
+    students_can_submit_proposals = models.BooleanField()
+    student_withdrawals_trigger_email = models.BooleanField()
+    enrollment_forms_can_be_re_submitted = models.BooleanField()
+    tax_forms_can_be_reviewed = models.BooleanField()
+    first_eval_emails_can_be_sent = models.BooleanField()
+    second_review_status_visible = models.BooleanField()
+    project_decisions_can_be_finalized = models.BooleanField()
+    student_home_address_is_editable = models.BooleanField()
+    program = models.ForeignKey(Program, on_delete=models.CASCADE)
```
sks444: > The Feature set are defined as part of the Program ; 1-1 relationship by the look of things. Needs a foriegn key here.

I already have a foreign key here. Is it wrong?

gsoc/models.py
jayvdb: the bool doesnt need to exist; it can be derived.

gsoc/models.py
jayvdb: as a 1-1 , it should be at the top, as it is also the primary key

requirements.txt
```diff
@@ -8,3 +8,4 @@ pillow
 ruamel.yaml
 markdown2
 python_dateutil
+django-eventtools
```
jayvdb: put this with the other django req, insert roughly in alpha order

gsoc/models.py
jayvdb: rename to `Features`

sks444: @jayvdb, Django will automatically add a 's' in the of the model names.

hemangsk:
LGTM, Please resolve the conflicts :)

jayvdb:
Please never use GitHub's stupid rebase button.  Now you have two commits.  Only one commit is acceptable.

nkprince007:
@gitmate-bot rebase

jayvdb:
ack https://github.com/coala/community/pull/64/commits/3cf3117c1a97f43ca0d7f537a492c478460e01ba

jayvdb:
@gitmate-bot ff

Issue #57, opened 2017-12-18
GCI: Create RSS feed of new tasks


jayvdb:
http://community.coala.io/static/tasks.yaml contains all tasks, sorted by task identifier.

Create an RSS feed of tasks, sorted by `last_modified`

jayvdb:
Example of our first published list of new tasks https://github.com/coala/community/commit/78c4c92341be8c73679f0180de63b3756fe86199#diff-feb0e78c555f43ae0c7b29fd3d81dbc2

The RSS feed should point to the task URL if available(usually it is a public issue), as well as the GCI task.

jayvdb:
Note that this task does not require detecting 'new tasks'.  The `tasks.yaml` is already loaded during the Travis CI build; it just needs to be converted to RSS and written to the deploy directory.  If the RSS is correctly written, RSS readers will automatically highlight new items appearing in the feed.

PR #65, opened 2017-12-20
gci/feeds: Create RSS feed of tasks

nalinbhardwaj:
Creates RSS feed for tasks. Task descriptions are rendered HTML from markdown, along with external URL, if any. Link points to GCI task link.

Closes https://github.com/coala/community/issues/57

### DIFF
--- a/community/settings.py
+++ b/community/settings.py
@@ -25,7 +25,7 @@ SECRET_KEY = 'v-#^6q@r4@lu*o%1$*c^7fpt8nl__*ac&oc1-of6)cwkdsp0#1'
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 
-ALLOWED_HOSTS = []
+ALLOWED_HOSTS = ['127.0.0.1', 'localhost', 'testserver']
 
 
 # Application definition
--- a/community/urls.py
+++ b/community/urls.py
@@ -8,6 +8,7 @@ from django.conf import settings
 from django.views.generic import TemplateView
 
 from gci.views import index as gci_index
+from gci.feeds import LatestTasksFeed as gci_tasks_rss
 from activity.scraper import activity_json
 from twitter.view_twitter import index as twitter_index
 
@@ -36,6 +37,12 @@ urlpatterns = [
         name='activity',
         distill_func=get_index,
         distill_file='activity/index.html',
+    ),
+    distill_url(
+        r'gci/tasks/rss.xml', gci_tasks_rss(),
+        name='gci-tasks-rss',
+        distill_func=get_index,
+        distill_file='gci/tasks/rss.xml',
     ),
     distill_url(
         r'gci/', gci_index,
--- a/gci/feeds.py
+++ b/gci/feeds.py
@@ -1,0 +1,49 @@
+import os.path
+from ruamel.yaml import YAML
+import markdown2
+import dateutil.parser
+
+from django.contrib.syndication.views import Feed
+from community.git import get_deploy_url, get_owner
+
+
+class LatestTasksFeed(Feed):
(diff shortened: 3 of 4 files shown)


community/settings.py
```diff
@@ -25,7 +25,7 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 
-ALLOWED_HOSTS = []
+ALLOWED_HOSTS = ['127.0.0.1', 'localhost', 'testserver']
```
hemangsk: Unrelated change?

nalinbhardwaj: No, the build will fail because the syndication feed framework needs `testserver` to be allowed for `django_distill` builds.

community/urls.py
```diff
@@ -37,6 +38,12 @@ def get_index():
         distill_func=get_index,
         distill_file='activity/index.html',
     ),
+    distill_url(
+        r'gci/rss/', gci_rss(),
+        name='gci-rss',
+        distill_func=get_index,
+        distill_file='gci/rss/index.xml',
```
jayvdb: `gci/tasks/rss.xml` ?

community/settings.py
hemangsk: Okay 👍

.ci/build.sh
```diff
@@ -9,7 +9,7 @@ if [[ -n "$GCI_TOKEN" ]]; then
   python manage.py cleanse_gci_task_data private _site
   rm -rf private/
 else
-  python manage.py fetch_old_gci_task_data _site || true
+  python manage.py fetch_old_gci_task_data private || true
```
jayvdb: oh, dont touch this :P

community/urls.py
nalinbhardwaj: `index.xml` is better since then you don't have to type the whole thing, how about `gci/tasks-rss/` or something then?

gci/feeds.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class LatestTasksFeed(Feed):
+    title = 'GCI tasks feed'
+    link = get_deploy_url() + '/gci/rss/'
+    description = 'GCI tasks ordered by modification time.'
+
+    def items(self):
+        yaml = YAML()
+        with open(os.path.join('private', 'tasks.yaml')) as f:
```
jayvdb: you need to load `_site/tasks.yaml`.  `private/tasks.yaml` is data which is private and can not be published.

.ci/build.sh
nalinbhardwaj: Builds won't work without API token then?

gci/feeds.py
```diff
@@ -0,0 +1,34 @@
...
+
+        return res
+
+    def item_title(self, item):
+        return item['name']
+
+    def item_description(self, item):
+        desc = item['description']
+        if item['external_url']:
+            desc += '\n\nExternal URL: [' + item['external_url'] + \
```
jayvdb: all this string appending is a bit ugly.  Could you use `%s` or `{url}`

community/urls.py
```diff
@@ -37,6 +38,12 @@ def get_index():
         distill_func=get_index,
         distill_file='activity/index.html',
     ),
+    distill_url(
+        r'gci/rss/', gci_rss(),
```
jayvdb: this and the next line also need to be updated.

community/urls.py
```diff
@@ -8,6 +8,7 @@
 from django.views.generic import TemplateView
 
 from gci.views import index as gci_index
+from gci.feeds import LatestTasksFeed as gci_rss
```
jayvdb: gci_tasks_rss

Issue #32, opened 2017-12-09
Import coala.io contributors functionality


jayvdb:
https://coala.io/#/getinvolved is part of https://github.com/coala/landing-frontend , and it uses data collated from the webservice provided by https://gitlab.com/coala/landing .

The same backend data should be imported into a data model in this repo, with a similar UI, but UI is not so important.

Note: The data from the webservice is a bit out of date.  Not a problem for this task.

sks444:
@jayvdb, One way to do it is to get all the contributor's data into a JSON file and load it into a `Contributor` model by `loaddata` but for this, we need a database and we are not using one here.  So how should we do it another way?

ashwini0529:
@sks444 contributors dataset is something that will be updated. So I would suggest to work with some sort of API that could fetch contributors and show them.

sks444:
@ashwini0529,
 

> contributors dataset is something that will be updated. So I would suggest to work with some sort of API that could fetch contributors and show them.

Yes, that is what I was thinking. Thank you :)
BTW, how we are fetching the data in https://github.com/coala/landing-frontend from the backend?

sks444:
I am thinking about making an API which scrapes the data from https://coala.io/#/getinvolved and posts it in the Contributor model.

ashwini0529:
@sks444 why do you need to store the data?

sks444:
@ashwini0529, Because that is what the issue is about, we need to import the data into a data model, we are not showing it here, yet.

sks444:
@ashwini0529, Oh, I have got it now. We don't have to do all this, we just have to implement what already has been done in https://gitlab.com/coala/landing/tree/master/backend/org for this repo, as @jayvdb, had said earlier, this task is a 'copy and paste.'
@yashraj-Dhanuka, are you working on it or should I create a PR?

yashraj-Dhanuka:
@sks444 Actually my exams were going on, so i could not get time on this.
I will work on it today , Thanks for the help bro.

jayvdb:
> we just have to implement what already has been done in gitlab.com/coala/landing/tree/master/backend/org for this repo

Not quite.  This repository should not re-create all that data.  It is time consuming.  This repository should fetch that data from the *webservice* provided by http://gitlab.com/coala/landing .

To do this task, you must run http://gitlab.com/coala/landing on your machine to use for testing, populate the data in that app , and then use its webservice (web api on your machine) to populate a new data model for this repository.

[...]

PR #70, opened 2018-01-07
Import coala.io contributors functionality

sks444:
Closes https://github.com/coala/community/issues/32


data/management/commands/import_contributors_data.py
```diff
@@ -0,0 +1,51 @@
...
+                reviews=reviews
+            )
+            if created:
+                c.save()
+                display_format = "\nContributor, {}, has been saved."
+                print(display_format.format(c))
+        except Exception as ex:
+            print(str(ex))
+            msg = "\n\nSomething went wrong saving this contributor: {}\n{}".format(login, str(ex))
+            print(msg)
```
andrewda: Any reason to save this in a `msg` variable first before printing? Just dilutes the code in my opinion.

data/management/commands/import_contributors_data.py
```diff
@@ -0,0 +1,51 @@
...
+                name=name,
+                bio=bio,
+                num_commits=num_commits,
+                issues_opened=issues_opened,
+                reviews=reviews
+            )
+            if created:
+                c.save()
+                display_format = "\nContributor, {}, has been saved."
+                print(display_format.format(c))
```
andrewda: Same as below - probably don't need to save the message in a variable first.

data/management/commands/import_contributors_data.py
```diff
@@ -0,0 +1,51 @@
...
+        except Exception as ex:
+            print(str(ex))
```
andrewda: We're printing the exception below, right? No need to print it here.

data/management/commands/import_contributors_data.py
```diff
@@ -0,0 +1,48 @@
+import requests
+import json
+from data.models import Contributor
+from django.core.management.base import BaseCommand
+
+IMPORT_URL = 'https://webservices.coala.io/contrib/'
```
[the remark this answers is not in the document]
jayvdb: this needs to be moved to an environment variable, so that it is not coala specific.

jayvdb:
The data import Command needs to be run from Travis

community/urls.py
```diff
@@ -63,4 +64,6 @@ def get_index():
         distill_func=get_index,
         distill_file='log/index.html',
     ),
+    url(r'^data/', include('data.urls')),
```
jayvdb: you need to use `distill_url`  ;-)

sks444: damn! never used that, let me read their docs then :)

jayvdb:
@gitmate-bot ff

Issue #33, opened 2017-12-09
Import OpenHub data


jayvdb:
See https://www.openhub.net/orgs/coala/ , and the OpenHub apis.  We need an OpenHub app that collects all of the data they make available for use by this website.

PR #77, opened 2018-02-17
Import OpenHub data

sks444:
Closes https://github.com/coala/community/issues/33
Here is what has been done till now:
1. Imported [portfolio projects ](https://github.com/blackducksoftware/ohloh_api/blob/master/reference/portfolio_projects.md)data for the organization, who load this app
2. Imported [outside projects](https://github.com/blackducksoftware/ohloh_api/blob/master/reference/outside_projects.md) data for the organization who load this app
3. Imported [outside committers](https://github.com/blackducksoftware/ohloh_api/blob/master/reference/outside_committers.md) data for the organization who load this app 
4. Imported [affiliated committers](https://github.com/blackducksoftware/ohloh_api/blob/master/reference/affiliated_committers.md) data for the organization who load this app
5. Imported [orgs](https://github.com/blackducksoftware/ohloh_api/blob/master/reference/organization.md) data for the organization who load this app 
see https://github.com/blackducksoftware/ohloh_api


sks444:
For reviewing the model you can take a look at the json data here https://dpaste.de/57iY

jayvdb:
Please actually import the data into the data model.
I'll probably need to add a key to netlify & travis , etc so that your import process works.

sks444:
@jayvdb, I am little bit confused here, since OpenHub APIs provides XML data so I was trying to use this [library](https://github.com/lev-veshnyakov/django-import-data) which directly import XML data into Django model from a url but the problem is for this we have to write XSLT file and also I am having some trouble with one to one fields. So, I just wanted to know, is there any library for this or should I go on with this one?

sks444:
@jayvdb , data imported :)

.ci/build.sh
```diff
@@ -12,5 +12,11 @@ else
   python manage.py fetch_old_gci_task_data _site || true
 fi
 
+python manage.py import_organization_data
```
jayvdb: create a single command to invoke all the other commands, so that the python runtime doesnt need to start so many times.

community/settings.py
```diff
@@ -122,3 +123,4 @@
 MEDIA_URL = '/media/'
 MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
 STATICFILES_DIRS = ['static/']
+openhub_api_key = 'enter_your_key'
```
jayvdb: should be upper case, and named like the others, so maybe `OH_TOKEN`

openhub/apps.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.apps import AppConfig
+
+
+class OpenhubConfig(AppConfig):
```
jayvdb: `OpenHub*` ?

sks444: @jayvdb, Actually the name of the app is `openhub` and here it automatically converts the first letter to uppercase.

community/settings.py
```diff
@@ -131,3 +132,4 @@
 MEDIA_URL = '/media/'
 MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
 STATICFILES_DIRS = ['static/']
+OH_TOKEN = ''
```
jayvdb: you need to get the value from the environment variable `OH_TOKEN`.  Use `gci.client.get_api_key('OH')`

openhub/views.py
```diff
@@ -0,0 +1,7 @@
+from openhub.models import OpenhubOrganization
+from django.shortcuts import render
+
+def index(request):
+	openhub_organizations = OpenhubOrganization.objects.all()
+	args = {"openhub_organizations": openhub_organizations}
+	return render(request, 'openhub_organizations.html', args)
```
jayvdb: EOL at EOF

openhub/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.conf.urls import url
+
+from . import views
+
+urlpatterns = [
+    url(r'^$', views.index, name='index'),
+]
```
jayvdb: EOL at EOF

community/settings.py
```diff
@@ -11,7 +11,7 @@
 """
 
 import os
-
+from gci.config import get_api_key
```
jayvdb: this causes a cyclic import.

you could move the OH_TOKEN constant into the openhub app, which is the only place it is used.

sks444: Thanks :)

jayvdb:
It looks like it is importing multiple orgs; it should only import coala.

and there is this in the Netlify log

```
2:57:14 AM: Something went wrong saving this OpenhubOrganization:
2:57:14 AM: local variable 'name' referenced before assignment
2:57:14 AM: All OpenHub data is imported
```

templates/index.html
```diff
@@ -7,5 +7,6 @@
     <li><a href="/twitter/">Twitter Feed</a>
     <li><a href="/log/">Logs</a>
     <li><a href="/contributors/">Contributors data</a>
+    <li><a href="/openhub_organizations/">OpenHub Organizations</a>
```
jayvdb: no, we only need the coala org initially.  so just `/openhub/` , and list the projects in OpenHub under the "coala" org (where coala is determined by the repo owner name).

there is some fun analytics we can do linking to other orgs, but importing 'all orgs' is too much data; it takes too long, and Netlify kills the job

sks444:
>It looks like it is importing multiple orgs; it should only import coala.

No, actually we have a API which can import all the organizations of OpenHub
We are also importing all the data of coala (the org who load this app) see https://github.com/coala/community/pull/77#issue-169765167

sks444:
>and there is this in the Netlify log

I am fixing the Netlify, since we are getting OH_TOKEN here. But it seems we are not getting it in travis

openhub/management/commands/import_openhub_data.py
```diff
@@ -0,0 +1,15 @@
+from django.core.management import BaseCommand
+from django.core.management import call_command
+
+
+class Command(BaseCommand):
+	help = "This command invoke all the importing data command"
+	call_command('import_affiliated_committers_data')
```
jayvdb: In Travis, the environment variables can not be exposed to Pull Requests, for security reasons.

As a result, this command needs to check if OH_TOKEN can be obtained.  If there is an exception, catch it, emit a note, and do nothing.

community/settings.py
```diff
@@ -11,7 +11,6 @@
 """
 
 import os
-
```
jayvdb: this blank line is necessary

openhub/management/commands/import_affiliated_committers_data.py
```diff
@@ -0,0 +1,71 @@
+import sys
+import json
+import requests
```
jayvdb: blank line separates std libs and custom packages

openhub/management/commands/import_affiliated_committers_data.py
```diff
@@ -0,0 +1,71 @@
...
+                name = name,
+                org = org,
+                kudos = kudos,
+                level = level,
+                most_commits = cr1,
+                most_recent_commit = cr2
+                )
+            if created:
+                c.save()
+                print ("\nAffiliatedCommitter, {}, has been saved.".format(c))
```
jayvdb: use `logging`

openhub/management/commands/import_affiliated_committers_data.py
```diff
@@ -0,0 +1,71 @@
...
+        except Exception as ex:
+            print ("\n\nSomething went wrong saving this AffiliatedCommitter: {}\n{}".format(name, str(ex)))
+
+
+    def handle(self, *args, **options):
+        """
+        Makes a GET request to the  API.
```
jayvdb: `  `

openhub/management/commands/import_affiliated_committers_data.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        n = 100
+        for i in range(1, n):
+            IMPORT_URL = 'https://www.openhub.net/orgs/'+org_name+'/affiliated_committers.xml?api_key='+OH_TOKEN+'&page='+str(i)
```
jayvdb: this isnt a constant, so shouldnt be uppercase

openhub/management/commands/import_outside_committers_data.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+org_name = get_owner()
+class Command(BaseCommand):
+    def import_data(self, contributor):
+        name = contributor.get('name', None)
+        kudos = contributor.get('kudos', None)
+        level = contributor.get('level', None)
+        affiliated_with = contributor.get('affiliated_with', None)
+        org = org_name
```
jayvdb: no need for variable name `org`; `org_name` is better.

openhub/management/commands/import_outside_committers_data.py
```diff
@@ -0,0 +1,66 @@
...
+
+from openhub.models import OutsideCommitter, ContributionsToPortfolioProject
+from django.core.management.base import BaseCommand
+from community.git import get_owner
+from openhub.oh_token import OH_TOKEN
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+org_name = get_owner()
```
jayvdb: not helpful to put this at the module level.

instead, add caching in the `get_owner` function, so that it can be called many times without performance problems.

openhub/management/commands/import_outside_committers_data.py
```diff
@@ -0,0 +1,66 @@
...
+class Command(BaseCommand):
+    def import_data(self, contributor):
```
jayvdb: this doesnt use `self`.  It shouldnt be in the class.

openhub/management/commands/import_outside_committers_data.py
```diff
@@ -0,0 +1,66 @@
...
+                logging.debug("AffiliatedCommitter {} has been saved".format(name))
+        except Exception as ex:
+            logging.debug("\n\nSomething went wrong saving this OpenhubOrganization: {}\n{}".format(name, str(ex)))
+
+    def handle(self, *args, **options):
+        """
+        Makes a GET request to the API.
+        """
+        n=100
+        for i in range(1, n):
```
jayvdb: the logic should be in the app modules, not in the command.

openhub/management/commands/import_portfolio_projects_data.py
```diff
@@ -0,0 +1,74 @@
...
+        n=100
+        for i in range(1, n):
+            import_url = 'https://www.openhub.net/orgs/'+org_name+'/projects.xml?api_key='+OH_TOKEN+'&page='+str(i)
+            resp = requests.get(import_url)
+            try:
+                jsonString = json.dumps(xmltodict.parse(resp.content), indent=4)
+                json_object = json.loads(jsonString)
+                data = json_object['response']['result']['portfolio_projects']['project']
+            except:
+                break
```
jayvdb: use logging for all exceptions

openhub/management/commands/import_outside_committers_data.py
sks444: @jayvdb, when I am removing the `self` from the `import_data()` method, I am getting this error:
```
TypeError: import_data() takes 1 positional argument but 2 were given
```

sks444: Got it; I have to use `@staticmethod` decorator :+1:

openhub/management/commands/import_affiliated_committers_data.py
```diff
@@ -0,0 +1,74 @@
...
+
+from community.git import get_owner
+from openhub.oh_token import OH_TOKEN
+from django.core.management.base import BaseCommand
+from openhub.models import MostCommit, MostRecentCommit, AffiliatedCommitter
+
+logging.basicConfig(level=logging.DEBUG)
+
+n = 100
+for i in range(1, 2):
```
jayvdb: this is effectively invalid syntax.  You replace the module level `Command` on the second iteration

no, the logic, all of it, needs to be modules in `openhub/` .  The `Command` *must* be at the top level of this module, not in any `if` or loop, and must only invoke the logic in `openhub/` modules (not subdirectories under that).  See the GCI app `Command` for an example.

openhub/management/commands/import_openhub_data.py
```diff
@@ -0,0 +1,21 @@
...
+from django.core.management import call_command
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+class Command(BaseCommand):
+	help = "This command invoke all the importing data command"
+	try:
+		call_command('import_affiliated_committers_data')
+		call_command('import_organization_data')
```
jayvdb: why import affiliated committers before the org?  that doesnt seem right.

openhub/management/commands/import_openhub_data.py
```diff
@@ -0,0 +1,21 @@
...
+		call_command('import_outside_committers_data')
+		call_command('import_outside_projects_data')
+		call_command('import_portfolio_projects_data')
+	except Exception as ex:
+		logging.debug(str(ex))
+		pass
+
+	def handle(self, *args, **options):
+		self.stdout.write("All OpenHub data is imported")
```
jayvdb: use `logging`.  No prints, no stdout/stderr.  See our `log` app for why.

openhub/management/commands/import_affiliated_committers_data.py
```diff
@@ -0,0 +1,74 @@
...
+    resp = requests.get(import_url)
+    jsonString = json.dumps(xmltodict.parse(resp.content), indent=4)
+    json_object = json.loads(jsonString)
+    try:
+        data = json_object['response']['result']['affiliated_committers']['affiliator']
+    except Exception as ex:
+        logging.debug(ex)
+        break
+    class Command(BaseCommand):
+        @staticmethod
```
jayvdb: If it doesnt use `self`, it shouldnt be inside the class.  `staticmethod` usually means you've designed it wrong.

openhub/management/commands/import_affiliated_committers_data.py
```diff
@@ -0,0 +1,53 @@
...
+            if created:
+                c.save()
+                logging.debug('\nAffiliatedCommitter {} has been saved'.format(name))
+        except Exception as ex:
+            logging.debug('\n\nSomething went wrong saving this OpenhubOrganization: {}\n{}'
+                .format(name, str(ex)))
+
+    def handle(self, *args, **options):
+        for affiliator in get_affiliated_committers():
+            self.import_data(affiliator)
```
sks444: @jayvdb, we are making use of self here, right? Inside the class.

openhub/management/commands/import_organization_data.py
```diff
@@ -0,0 +1,85 @@
...
+            'infographic_details']['affiliators_committing_to_outside_projects']
+
+        try:
+            cr, create = InfographicDetail.objects.get_or_create(
+                outside_committers=outside_committers,
+                outside_committers_commits=outside_committers_commits,
+                projects_having_outside_commits=projects_having_outside_commits,
+                portfolio_projects=portfolio_projects,
+                affiliators=affiliators,
+                affiliators_committing_to_portfolio_projects=affiliators_committing_to_portfolio_projects,
```
sks444: @jayvdb, How one should fix this? 
Lets say I am doing this
```
affiliators_committing_to_portfolio_projects = \
    affiliators_committing_to_portfolio_projects,
```
Then again when I am running coala it fixes this to this:
```
affiliators_committing_to_portfolio_projects=affiliators_committing_to_portfolio_projects,
```
and then showing the line length error

openhub/management/commands/import_openhub_data.py
```diff
@@ -0,0 +1,23 @@
...
+from django.core.management import BaseCommand
+from django.core.management import call_command
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+class Command(BaseCommand):
+    help = 'This command invoke all the importing data command'
+
+    try:
```
jayvdb: this should be in `handle`.  instantiating a Command shouldnt run it automatically.

openhub/portfolio_projects.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def import_data(project):
+    name = project.get('name', None)
+    activity = project.get('activity', None)
+    i_use_this = project.get('i_use_this', None)
+    community_rating = project.get('community_rating', None)
+    org = get_owner()
+    commits = project.get(
+        'twelve_mo_activity_and_year_on_year_change', None).get('commits', None)
```
jayvdb: why does this `get` default to `None`, and then does another `get` on the result?
If it is `None`, the second get will fail.

Also you repeat `project.get('twelve_mo_activity_and_year_on_year_change', None)` lots of times.  Get it once, put it in a variable, and then use the variable.

openhub/portfolio_projects.py
```diff
@@ -0,0 +1,80 @@
...
+    change_in_contributors = project.get(
+        'twelve_mo_activity_and_year_on_year_change', None).get(
+        'change_in_contributors', None)
+    percentage_change_in_committers = project.get(
+        'twelve_mo_activity_and_year_on_year_change', None).get(
+        'percentage_change_in_committers', None)
+
+    try:
+        (cr, create) = PortfolioProjectActivity.objects.get_or_create(
+            commits=commits,
```
jayvdb: you can probably do `**project['twelve_mo_activity_and_year_on_year_change']`

that will auto fill all these fields, as the field names are the same as the dict member names

openhub/outside_projects.py
```diff
@@ -0,0 +1,57 @@
...
+    claimed_by = project.get('claimed_by', None)
+    i_use_this = project.get('i_use_this', None)
+    community_rating = project.get('community_rating', None)
+    affiliates_contributing = project.get('affiliates_contributing', None)
+    commits_by_current_affiliates = project.get(
+        'commits_by_current_affiliates', None)
+
+    try:
+        (c, created) = OutsideProject.objects.get_or_create(
+            name=name,
```
jayvdb: use `**` trick here and elsewhere

openhub/outside_committers.py
```diff
@@ -0,0 +1,64 @@
...
+from community.git import get_owner
+from openhub.oh_token import OH_TOKEN
+from openhub.models import OutsideCommitter, ContributionsToPortfolioProject
+
+
+def get_outside_committers_data():
+    n = 100
+    data1 = []
+    for i in range(1, n):
+        import_url = 'https://www.openhub.net/orgs/' + get_owner() \
```
jayvdb: create a function to build this URL, as the URL is very similar for each of them.

Also the page iteration process is very similar, so should be possible to create a generic function to get the data into a JSON like dict .

Also, looks like you could use https://github.com/jpopelka/openhubinfo to fetch the data as XML.  You probably need to add 'org' support to that script.  We can use your fork of that script to create JSON files, and then use your existing code code to import those json files.

openhub/affiliated_committers.py
```diff
@@ -0,0 +1,46 @@
+import logging
+
+from community.git import get_owner
+from openhub.models import MostCommit, MostRecentCommit, AffiliatedCommitter
+
+logging.basicConfig(level=logging.DEBUG)
```
jayvdb: why is this here?

openhub/data.py
```diff
@@ -0,0 +1,53 @@
...
+from openhub.portfolio_projects import get_portfolio_projects_data
+from openhub.outside_committers import get_outside_committers_data
+from openhub.affiliated_committers import get_affiliated_committers_data
+
+
+def get_data(for_what):
+    n = 100
+    data1 = []
+    for i in range(1, n):
+        import_url = 'https://www.openhub.net/orgs/' + get_owner() \
```
jayvdb: please do not use `\` , ever.  it is an abomination.

openhub/management/commands/import_portfolio_projects_data.py
```diff
@@ -0,0 +1,12 @@
+from openhub.data import get_data
+from openhub.portfolio_projects import import_data
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = 'Import PortfolioProjects data'
+
+    def handle(self, *args, **options):
+        data = 'projects'
```
jayvdb: put this string into a class level CONSTANT with a more meaningful variable name

openhub/organization.py
```diff
@@ -0,0 +1,61 @@
...
+
+    try:
+        (cr, create) = InfographicDetail.objects.get_or_create(
+            **org['infographic_details']
+            )
+        if create:
+            cr.save()
+        (c, created) = Organization.objects.get_or_create(
+            infographic_details=cr,
+            name=name,
```
jayvdb: you can still use `**` here

openhub/outside_committers.py
```diff
@@ -0,0 +1,39 @@
...
+    org = get_owner()
+
+    try:
+        (cr, create) = ContributionsToPortfolioProject.objects.get_or_create(
+            **contributor['contributions_to_portfolio_projects'])
+        if create:
+            cr.save()
+        (c, created) = OutsideCommitter.objects.get_or_create(
+            contributions_to_portfolio_projects=cr,
+            name=name,
```
jayvdb: `**`

openhub/portfolio_projects.py
```diff
@@ -0,0 +1,40 @@
...
+
+    try:
+        (cr, create) = PortfolioProjectActivity.objects.get_or_create(
+            **project['twelve_mo_activity_and_year_on_year_change']
+            )
+        if create:
+            cr.save()
+        (c, created) = PortfolioProject.objects.get_or_create(
+            twelve_mo_activity_and_year_on_year_change=cr,
+            name=name,
```
jayvdb: `**`

openhub/affiliated_committers.py
```diff
@@ -0,0 +1,46 @@
...
+            )
+        if create1:
+            cr1.save()
+        (cr2, create2) = MostRecentCommit.objects.get_or_create(
+            **affiliator['most_recent_commit']
+            )
+        if create2:
+            cr2.save()
+        (c, created) = AffiliatedCommitter.objects.get_or_create(
+            name=name,
```
jayvdb: use `**`, but you need to modify `affiliator` first.  e.g. `affiliator['most_commits'] = cr1`

openhub/organization.py
sks444: can you give me a quick example?

openhub/affiliated_committers.py
sks444: I was not able to use `**` here because `org` field is not into `dict`

openhub/outside_committers.py
sks444: `org` field problem

openhub/portfolio_projects.py
sks444: I have used `**` everywhere and its working fine but don't why that is not working in this case, I have tried every possible way.

openhub/management/commands/import_affiliated_committers_data.py
```diff
@@ -0,0 +1,13 @@
+from openhub.data import get_data
+from django.core.management.base import BaseCommand
+from openhub.affiliated_committers import import_data
+
+
+class Command(BaseCommand):
+    help = 'Import AffiliatorCommitters data'
+
+    affiliated_committers = 'affiliated_committers'
```
jayvdb: use a CONSTANT name, common in all of these Command, so that these Command `handle` method is identical.

You could, and should, create a `BaseCommand` which they all inherit `handle` from.

openhub/management/commands/import_organization_data.py
```diff
@@ -0,0 +1,12 @@
...
+from openhub.organization import get_organization_data, import_data
+
+
+class Command(BaseCommand):
+    help = 'Import Organization data'
+
+    org = get_organization_data()
+
+    def handle(self, *args, **options):
+
```
jayvdb: remove

openhub/oh_token.py
```diff
@@ -0,0 +1,8 @@
+import logging
+
+from gci.config import get_api_key
+
+try:
+    OH_TOKEN = get_api_key('OH')
+except Exception as ex:
+    logging.debug('OH_TOKEN can not be obtained: \n{}'.format(str(ex)))
```
jayvdb: also set `OH_TOKEN = None` , otherwise this module has an inconsistent set of members.

templates/openhub.html
```diff
@@ -0,0 +1,48 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+
+    <!-- Bootstrap CSS -->
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/css/bootstrap.min.css" integrity="sha384-Gn5384xqQ1aoWXA+058RXPxPg6fy4IWvTNh0E263XmFcJlSAwiGgFAW/dAiS6JXm" crossorigin="anonymous">
+
+    <title>OpenHub Organization Data</title>
+  </head>
+  <body>
+    <h1>All of our Portfolio Projects</h1>
+<ul>
```
jayvdb: `ul` ?

templates/openhub.html
```diff
@@ -0,0 +1,48 @@
...
+    </div>
+  </div>
+</div>
+</div>
+<hr>
+     {% endfor %}
+</ul>
+    <!-- Optional JavaScript -->
+    <!-- jQuery first, then Popper.js, then Bootstrap JS -->
+    <script src="https://code.jquery.com/jquery-3.2.1.slim.min.js" integrity="sha384-KJ3o2DKtIkvYIK3UENzmM7KCkRr/rE9/Qpg6aAZGJwFDMVNA/GpGFF93hXpG5KkN" crossorigin="anonymous"></script>
```
jayvdb: why do you need jquery or popper?

openhub/management/commands/import_organization_data.py
sks444: Actually, this is the only command which doesn't have identical `handle` with other.

openhub/management/commands/import_affiliated_committers_data.py
sks444: >You could, and should, create a BaseCommand which they all inherit handle from.

@jayvdb, not getting it properly. Can you explain it a little bit more?

jayvdb: CONSTANTS should be in CAPS.

Create class CONSTANTs for the two things which are different between `import_affiliated_committers_data.py`, `import_portfolio_projects_data.py`, etc.

Dont worry about the `BaseCommand` bit for now.

openhub/management/commands/import_organization_data.py
jayvdb: remove the blank line

openhub/management/commands/import_affiliated_committers_data.py
jayvdb: DO NOT create class called `CONSTANTS`.  Read what reviewers say carefully.

Each of these `Command` classes should have only four members, 'help', 'DATA_NAME' and 'IMPORT_FUNC' and `handle`, and the `handle` must be _identical_ in all of them.

openhub/management/commands/import_affiliated_committers_data.py
```diff
@@ -0,0 +1,14 @@
+from openhub.data import get_data
+from django.core.management.base import BaseCommand
+from openhub.affiliated_committers import import_data
+
+
+class Command(BaseCommand):
+    help = 'Import AffiliatorCommitters data'
+
+    COllECTIONS = 'affiliated_committers'
```
jayvdb: `COllECTIONS` with lower case `l` ?

sks444: damn!

openhub/management/commands/import_organization_data.py
```diff
@@ -0,0 +1,11 @@
+from django.core.management.base import BaseCommand
+from openhub.organization import get_organization_data, import_data
+
+
+class Command(BaseCommand):
+    help = 'Import Organization data'
+
+    ORG = get_organization_data()
```
jayvdb: instantiating a class shouldnt do real work.  it should only do set up.

in this case, `get_organization_data` has all sorts of exceptions which could occur and wont be managed easily

.ci/build.sh
```diff
@@ -12,7 +12,9 @@ else
   python manage.py fetch_old_gci_task_data _site || true
 fi
 
+
```
jayvdb: why add a blank line here.

openhub/oh_token.py
```diff
@@ -0,0 +1,9 @@
+import logging
+
+from gci.config import get_api_key
+
+try:
+    OH_TOKEN = get_api_key('OH')
+except Exception as ex:
+    OH_TOKEN = None
+    logging.debug('OH_TOKEN can not be obtained: \n{}'.format(str(ex)))
```
jayvdb: gah; why is this 'debug' level?  this is the error which causes the app to not work.

jayvdb:
ack https://github.com/coala/community/pull/77/commits/d83f4be33e10c42e5b48ff7df41c9b282b7377ab

jayvdb:
@gitmate-bot ff

PR #91, opened 2018-03-11
Add: GetInactiveIssue

RaiVaibhav:
Add Inactive Issue

### DIFF
--- a/.coafile
+++ b/.coafile
@@ -1,8 +1,9 @@
 [all]
-files = *.py, community/**/*.py, gci/**/*.py, activity/*.py
-ignore = gci/client.py, gci/migrations/**, private/**, gsoc/migrations/**
+files = **.py, **.js, **.sh
+ignore = .git/**, **/__pycache__/**, gci/client.py, */migrations/**, private/*
 max_line_length = 80
 use_spaces = True
+preferred_quotation = '
 
 [all.whitespace]
 use_spaces = True
@@ -10,17 +11,22 @@ bears = SpaceConsistencyBear
 default_actions = *: ApplyPatchAction
 
 [all.python]
+files = **.py
 language = Python
-bears = QuotesBear, PyFlakesBear
-preferred_quotation = '
+
+[all.python.default]
+ignore += */models.py
+bears = PEP8Bear, PycodestylBear, PyFlakesBear, QuotesBear
 default_actions = *: ApplyPatchAction
 
-[all.autopep8]
-bears = PEP8Bear, PycodestyleBear
-default_actions = PEP8Bear: ApplyPatchAction
-
 [all.linelength]
+ignore += */models.py
 bears = LineLengthBear
+
+[all.python.models]
+files = */models.py
+bears = LineLengthBear, PycodestylBear, PyFlakesBear, QuotesBear
+max_line_length = 120
 
 [all.links]
 bears = InvalidLinkBear
@@ -36,11 +42,11 @@ files = *.sh
 bears = ShellCheckBear
 shell = bash
 
-[generalization]
+[all.generalization]
 # Do not allow the word "coala" to ensure the repository can be generalized out
 # for use by other organizations.
 files = **
-ignore = .git/**, org_name.txt, .coafile, requirements.txt, .travis.yml, LICENSE, public/**, _site/**
+ignore += org_name.txt, .coafile, requirements.txt, .travis.yml, LICENSE, public/**, _site/**
 bears = KeywordBear
 language = python 3
 keywords = coala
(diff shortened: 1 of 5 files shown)


InactiveIssue/GetInactiveIssue.py
```diff
@@ -0,0 +1,71 @@
...
+from datetime import date
+from django.http import HttpResponse
+from gci.config import get_api_key
+import requests,json
+
+try:
+    GH_TOKEN = get_api_key('GH')
+except:
+    print("Unable to get the token")
+    pass
```
jayvdb: this does nothing.

but also get rid of code at the module level.  that is always a bad idea.

InactiveIssue/GetInactiveIssue.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+
+def run(request):
+    g1 = Github(GH_TOKEN)
```
jayvdb: get the token here, and catch the exception, and exit the function.  do not try to run without a token.

InactiveIssue/GetInactiveIssue.py
```diff
@@ -0,0 +1,64 @@
+from github import Github
```
jayvdb: please re-order your imports to comply with PEP8.

also use snake_case for filenames

RaiVaibhav: In my opinion output is correct, I have to output those issue in which someone is assigned from more than 60 days ,so here two case happen ,first he is assigned and not shown any activity (if yes it will act directly to issue list) and if no then it will check the last commit on issue and check it's date ,if last commit was made 60 days earlier then it will add the issue to issue list.

RaiVaibhav: Currently It's running only on coala/community repo.

RaiVaibhav: If you want I can write a script script for any further use, but only problem is of `timeline of Github api` which still needs many changes

jayvdb: oh, can we have it run on all repos in the org?

InactiveIssue/Get_Inactive_Issue.py
```diff
@@ -0,0 +1,64 @@
...
+                        if calculated_days >= 60:
+                            # for checking last commit update
+                            issues_number_list.append(issue_no)
+                    else:
+                        issues_number_list.append(issue_no)
+                break
+    return issues_number_list
+
+
+def inactiveIssue(request):
```
jayvdb: function names must comply with PEP8: snake case.  no upper case.

filenames also must not be upper case.

and IMO call your app 'inactive_issues', based on only a little reading like https://stackoverflow.com/questions/3098681/is-there-a-naming-convention-for-django-apps

InactiveIssue/GetInactiveIssue.py
RaiVaibhav: Yes we can but problem is time and rate limit as we are only limited to 5000 limit (not totally sure), it take approx 17 minutes to run on my local machine on `coala/coala` repo :sweat_smile:

InactiveIssue/Get_Inactive_Issue.py
```diff
@@ -0,0 +1,64 @@
...
+    try:
+        GH_TOKEN = get_api_key('GH')
+    except:
+        return HttpResponse('[]')
+    g1 = Github(GH_TOKEN)
+    org = g1.get_organization('coala')  # Ignore KeywordBear
```
sks444: @RaiVaibhav, make it generic, like for any org who load this app. use `get_owner` method from `community.git` to get the name of that org

InactiveIssue/Get_Inactive_Issue.py
```diff
@@ -0,0 +1,64 @@
...
+    repo = org.get_repo('community')
```
sks444: After getting the `org_name` get the list of all of its repo and run your code for each repo to get_inactive issues; store it along with its repo name and then print it.

RaiVaibhav: @sks444  Currently there is a reason for not doing this.
https://github.com/coala/community/pull/91#discussion_r174018094

InactiveIssue/Get_Inactive_Issue.py
jayvdb: This is not negotiable.  you need to not use the word `coala`.  use `get_owner` , and luckily you can use the owner name as the repo name, which will fetch `coala/coala`.  That is good enough for the first version of this.

RaiVaibhav:
@gitmate-bot rebase

PR #109, opened 2018-03-17
.coafile: Fix spelling mistake

jayvdb:


hemangsk:
ack a02f456

jayvdb:
@gitmate-bot ff

inactive_issues/inactive_issues_scraper.py
```diff
@@ -47,7 +47,7 @@ def run(issues):
 def inactive_issues_json(request):
     try:
         GH_TOKEN = get_api_key('GH')
-    except:
+    except Exception:
```
Makman2: Separate commit? :)

jayvdb: I normally am happy to include minor fixes in the 'upgrade coafile' patch, so we can see what was problematic at the time.

Makman2: Ah I forgot that Pycodestyle gets now enabled because the typo is fixed... Nvm :+1:

Issue #117, opened 2018-04-06
Add HTMLLintBear to `.coafile`


sks444:
ref: https://github.com/coala/community/pull/115#discussion_r179777613

PR #122, opened 2018-04-10
.coafile: Add HTMLLintBear

sks444:
This adds the HTMLLintBear and its suggested changes.

Closes https://github.com/coala/community/issues/117

### DIFF
--- a/.coafile
+++ b/.coafile
@@ -31,6 +31,12 @@ max_line_length = 120
 [all.links]
 bears = InvalidLinkBear
 
+[html]
+bears = HTMLLintBear
+files = **.html
+htmllint_ignore = optional_tag, concerns_separation
+ignore = public/**
+
 [javascript]
 files = static/**/*.js
 bears = JSHintBear
--- a/templates/activity.html
+++ b/templates/activity.html
@@ -1,41 +1,39 @@
-<!doctype html>
+<!DOCTYPE html>
 <html>
-
-<head>
+  <head>
     <title>Community Activity</title>
-    <style>
-    canvas {
+      <style>
+        canvas {
         -moz-user-select: none;
         -webkit-user-select: none;
         -ms-user-select: none;
-    }
-    </style>
-    <meta http-equiv="content-type" content="text/html; charset=utf-8" />
-</head>
-
-<body>
+        }
+      </style>
+      <meta charset="utf-8">
+      <meta content="text/html">
+  </head>
+  <body>
     <h1>Community Activity</h1>
     <br>
     <br>
     <div style="width:75%;">
-        <canvas id="canvas"></canvas>
+      <canvas id="canvas"></canvas>
     </div>
-    <select id="chartType">
-        <option value="Month">Year</option>
-        <option value="Week">Month</option>
-        <option value="Day">Week</option>
+    <select id="chart-type">
+      <option value="Month">Year</option>
+      <option value="Week">Month</option>
+      <option value="Day">Week</option>
     </select>
     <br>
     <br>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.7.1/Chart.bundle.js"></script>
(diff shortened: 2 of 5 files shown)


.coafile
```diff
@@ -31,6 +31,12 @@ max_line_length = 120
 [all.links]
 bears = InvalidLinkBear
 
+[html]
+bears = HTMLLintBear
+files = **.html
+htmllint_ignore = http_equiv, entities, optional_tag, concerns_separation, names, protocol
```
nalinbhardwaj: Why are so many rules being ignored?

`http_equiv` seems fixable by replacing https://github.com/coala/community/blob/master/templates/activity.html#L13 by `<meta charset="utf-8">`

`entities` only triggers on https://github.com/coala/community/blob/master/templates/index.html#L11 which to me seems like a bug? I'm not sure why there is a carriage return there, maybe quotes were intended originally?

`protocol` is also fixable across the files.

I haven't manually checked others, but try to not ignore the rules.

sks444:
Thanks, @nalinbhardwaj, for the review.
Can you take a look again :)

templates/index.html
```diff
@@ -1,15 +1,15 @@
...
+  <ul>
     <li><a href="/gci/">Google Code-in</a>
     <li><a href="/activity/">GitHub activity</a>
     <li><a href="/twitter/">Twitter Feed</a>
     <li><a href="/info.txt">Build info</a>
     <li><a href="/log/">Logs</a>
     <li><a href="/contributors/">Contributors data</a>
-    <li><a title="List of all the issues on organization&#13;&#10;main repository on which assignee&#13;&#10;has not shown any activity&#13;&#10;for more than 2 months."
-           href="/static/inactive-issues.json">Inactive Issues</a>
+    <li><a title="List of all the issues on organization main repository on which assignee has not shown any activity for more than 2 months."
```
nalinbhardwaj: I think it should be `organization's`?

sks444:
@nalinbhardwaj, there seems to be some problem on breaking the commits into two. That bear will run on unchanged files and the ci will not pass.
 
What to you mean by this line?
>If you're going to keep a single commit, make it multi

nalinbhardwaj:
@sks444 : it should pass if you make the HTML changes before the coala config changes I think?

> What do you mean by this line?

By that, I just meant replacing `.coafile:` in the shortlog by `multi:` prefix, since the change spans multiple files.

sks444:
>it should pass if you make the HTML changes before the coala config changes I think?

Yeah, but doesn't sound good, right? At first, we are adding the bear then we are finding what changes to make. 

Going for the `multi:` then. :)

sks444:
@nalinbhardwaj, looks good?

templates/activity.html
```diff
@@ -1,41 +1,39 @@
...
-    }
-    </style>
-    <meta http-equiv="content-type" content="text/html; charset=utf-8" />
-</head>
-
-<body>
+        }
+      </style>
+      <meta charset="utf-8">
+      <meta charset="utf-8" content="text/html">
```
blazeu: This is probably not needed, 2 meta charset?

sks444: I think I should remove the line-12 meta tag?

templates/index.html
```diff
@@ -1,15 +1,15 @@
 <title>Community website</title>
```
jayvdb: why no `<html lang="en">` required?

sks444: No idea, why it was not added at the first place. I will add it 👍

sks444:
>split into two commits; one for tiding templates/: .. and another for .coafile: ...

Alright👍
Actually, we thought that it wouldn't be good to commit the changes suggested by bear before it.

jayvdb:
ack https://github.com/coala/community/pull/122/commits/812f613cd8d5b572f28d71316aebbecbd0e13bed https://github.com/coala/community/pull/122/commits/9261c9f6d0777a150e3ee69b146e4d7f52a4eb82

jayvdb:
@gitmate-bot ff

Issue #133, opened 2018-05-21
Set up meta_review app


li-boxuan:
Create meta_review app, complete template, and models.

This is part of GSoC'18 project meta-review.

See more about the proposed meta-review system: https://github.com/coala/cEPs/blob/master/cEP-0019.md

li-boxuan:
Please assign this issue to me. ;)

PR #134, opened 2018-05-21
Set up meta_review app

li-boxuan:
Build basic structure for the meta_review scoring system.
Complete templates/meta_review.html, meta_review/models.py.

Closes https://github.com/coala/community/issues/133


meta_review/models.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def __str__(self):
+        return 'Meta-reviewer: ' + self.login
+
+    class Meta:
+        ordering = ['rank']
+
+
+class Comment(models.Model):
+    id = models.TextField(default=None, primary_key=True)
```
sks444: Could you consider renaming `id` to `identifier` ? `id` would interfere with python's built-in function.

meta_review/models.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # number of negative reactions give away
+    neg_out = models.IntegerField(default=0, null=True)
+
+    # punishment points
+    punishment = models.FloatField(default=0, null=True)
+
+    # weight factor
+    weight_factor = models.FloatField(default=0.1, null=True)
+
```
hemangsk: Nice job on adding the comments!

templates/meta_review.html
```diff
@@ -0,0 +1,53 @@
+<!DOCTYPE html>
```
hemangsk: Can you post a screenshot of the generated screen?

li-boxuan: There's no 'UI' for that. The template is modified from templates/contributors.html, which looks like http://coala-community.netlify.com/contributors/

community/urls.py
```diff
@@ -127,6 +128,12 @@ def get_organization():
         distill_func=get_index,
         distill_file='contributors/index.html',
     ),
+    distill_url(
+        r'meta-review/$', meta_review_index,
```
sks444: IMO, keep it `meta_review`, as your app name is also the same.

community/urls.py
```diff
@@ -127,6 +128,12 @@ def get_organization():
...
+        name='meta-review-data',
```
sks444: meta_review_data, maybe? :)

community/urls.py
blazeu: snake_case is a bit unusual for url.

li-boxuan: As what we discussed here https://gitter.im/coala/community?at=5b0d5fe8b6eece791dc0de4e, snake case is a bit unusual for URL 😝 
Any voice is welcome :P

meta_review/views.py
```diff
@@ -0,0 +1,15 @@
...
+def index(request):
+    participants = Participant.objects.all().exclude(
+        Q(pos_in=0),
+        Q(neg_in=0),
+        Q(pos_out=0),
+        Q(neg_out=0),
+        Q(punishment=0)
+    )
+    args = {'participants': participants}
+    return render(request, 'meta_review.html', args)
```
sks444: Here you're rendering `meta_review.html`.

And [here](https://github.com/coala/community/pull/134/files#diff-f2db580a801ecfe0049dc81b9da19c6cR135) you're passing `meta-review/index.html` that could cause an error, IMO.

meta_review/models.py
li-boxuan: I would like to hear more voice.

I have thought it for a while and tried finding some evidence to support my original idea. I did found some, very weak though.

As https://docs.djangoproject.com/en/2.0/topics/db/models/#automatic-primary-key-fields points out,

By default, Django gives each model the following field:

> id = models.AutoField(primary_key=True)

Also, from https://github.com/django/django/blob/f1f4aeb22e7bc9b504f69f7cb111ac9bdedb5f1e/django/db/models/base.py#L1298-L1312 we can see that they allow the usage of `id` field as long as it has `primary_key=True`

meta_review/views.py
li-boxuan: Nope, it doesn't cause any error 😛 The name of template passed in to render method doesn't matter, as long as there is a corresponding file under template directory.

In this case, since `templates/meta_review.html` exists, `return render(request, 'meta_review.html', args)` is fine.

meta_review/models.py
blazeu: @hemangsk could you step in ;)

I'm pretty sure `id` is ok, but I'm not a django master.

hemangsk: It seems alright to me as well.

hemangsk:
ack 2929576

jayvdb:
@gitmate-bot ff

Issue #132, opened 2018-05-21
Move fetch_deployed_data out of gci app


li-boxuan:
https://github.com/coala/community/blob/e9d74a1b5ca293683b6cbe1f71757a03a52a57ac/gci/management/commands/fetch_deployed_data.py is reusable. It should not be in the gci app directory.

jayvdb:
Im not sure why, but it doesnt like being in the `community` app.

Maybe it looks in `INSTALLED_APPS`.
It isnt worth creating a new app for these generic commands, so I propose it be moved into the data app.

li-boxuan:
> Im not sure why, but it doesnt like being in the community app
I also tried that and it didn't work ;)

>  I propose it be moved into the data app.
Looks good to me.

PR #143, opened 2018-06-03
Implement meta-review scoring & ranking system

### LINKED ISSUE #142: Implement meta review scoring & ranking system
Data needs to be fetched from https://coala.github.io/gh-board/issues.json processed and stored in the database.

Reaction data needs to be analyzed for scoring and ranking.

This is part of GSoC'18 project meta-review.

See more about the proposed meta-review system: https://github.com/coala/cEPs/blob/master/cEP-0019.md

Please assign this issue to me. ;)

li-boxuan:
1. Data needs to be fetched from
   https://coala.github.io/gh-board/issues.json.
2. Reaction data needs to be analyzed for scoring and ranking.
3. Result needs to be displayed on the website and stored in the
   database.

Closes #142

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -4,19 +4,43 @@ set -e -x
 
 mkdir private _site public
 
-EXPORTED_DATA="static/tasks.yaml static/instances.yaml"
+META_REVIEW_DATA="meta_review.json"
+
+EXPORTED_DATA="static/tasks.yaml static/instances.yaml static/$META_REVIEW_DATA"
+
+ISSUES_JSON="issues.json"
+
+python manage.py fetch_deployed_data --allow-failure _site $EXPORTED_DATA
 
 if [[ -n "$GCI_TOKEN" ]]; then
   python manage.py fetch_gci_task_data private
   python manage.py cleanse_gci_task_data private _site
   rm -rf private/
-else
-  python manage.py fetch_deployed_data _site $EXPORTED_DATA
 fi
+
+# fetch deployed issues data in gh-board repo
+python manage.py fetch_deployed_data _site $ISSUES_JSON --repo-name gh-board
 
 python manage.py migrate
 python manage.py test
 python manage.py import_contributors_data
 python manage.py import_openhub_data
+
+if [[ -f "_site/$META_REVIEW_DATA" ]]; then
+  echo "File $META_REVIEW_DATA exists."
+  # Load meta_review data
+  python manage.py loaddata _site/$META_REVIEW_DATA
+else
+  echo "File $META_REVIEW_DATA does not exist."
+fi
+
+# Run meta review system
+python manage.py run_meta_review_system
+
+rm _site/$ISSUES_JSON
+
+# Dump meta_review data
+python manage.py dumpdata meta_review > _site/$META_REVIEW_DATA
+
 python manage.py collectstatic --noinput
 python manage.py distill-local public --force
--- a/community/git.py
+++ b/community/git.py
@@ -155,18 +155,23 @@ def get_upstream_repo():
     return parent
 
 
-def get_deploy_url():
-    """Obtain the http where deploys appear.
+def get_deploy_url(name=None):
+    """
+    Obtain the http where deploys appear.
+    When `name` is None, fetch current repo of current org,
+    otherwise, fetch a specified repo of current org.
     """
     # Use environment variable URL when Netlify detected
-    if os.environ.get('REPOSITORY_URL'):
+    if not name and os.environ.get('REPOSITORY_URL'):
         return os.environ.get('URL')
 
     url = get_remote_url()
     if url.resource != 'github.com':
         raise Exception('remotes %s is not supported' % url)
 
-    deploy_url = 'https://{url.owner}.github.io/{url.name}'.format(url=url)
+    owner = url.owner
+    path = name if name else url.name
+    deploy_url = 'https://%s.github.io/%s' % (owner, path)
 
     return deploy_url
 
--- a/data/management/commands/fetch_deployed_data.py
+++ b/data/management/commands/fetch_deployed_data.py
@@ -14,13 +14,20 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument('output_dir', nargs='?', type=str)
         parser.add_argument('filenames', nargs='+', type=str)
+        parser.add_argument('--repo-name', action='store',
+                            dest='repo-name', type=str)
+        parser.add_argument('--allow-failure', action='store_true',
+                            dest='allow-failure', default=False,
+                            help='Don\'t raise exceptions.')
 
     def handle(self, *args, **options):
         logger = logging.getLogger(__name__)
         output_dir = options.get('output_dir')
         filenames = options.get('filenames')
+        repo_name = options.get('repo-name')
+        allow_failure = options.get('allow-failure')
 
-        deploy_url = get_deploy_url()
+        deploy_url = get_deploy_url(repo_name)
         try:
             upstream_deploy_url = get_upstream_deploy_url()
         except RuntimeError as e:
@@ -34,9 +41,13 @@ class Command(BaseCommand):
             except Exception:
                 if upstream_deploy_url:
                     r = requests.get(upstream_deploy_url + '/' + filename)
-                    r.raise_for_status()
+                    if not allow_failure:
+                        r.raise_for_status()
                 else:
-                    raise
+                    if allow_failure:
+                        return
+                    else:
+                        raise
 
             filename = os.path.basename(filename)
             with open(os.path.join(output_dir, filename), 'wb') as f:
--- a/meta_review/dump_to_db.py
+++ b/meta_review/dump_to_db.py
@@ -1,0 +1,50 @@
+import logging
+
+from meta_review.models import Participant, Comment, Reaction
+
+
+def dump_all(participants, comments, reactions):
+    """
+    Dump all meta-review related data into database.
+    """
+    dump_participants(participants)
+    dump_comments(comments)
+    dump_reactions(reactions)
+
+
+def dump_data(cls, data):
+    """
+    Helper function for dumping data into database.
+    """
+    logger = logging.getLogger(__name__)
+    logger.info('dump %s data into database' % cls.__name__.lower())
+    try:
+        # there's no way to do bulk update, so we delete and then create
+        # bulk delete all data
+        cls.objects.all().delete()
+        # bulk create all data
+        cls.objects.bulk_create(data.values())
+    except Exception as ex:
+        logger.error('Something went wrong saving %s: %s'
+                     % (cls.__name__.lower(), ex))
+
+
+def dump_participants(participants):
+    """
+    Dump participants data into Django database.
+    """
+    dump_data(Participant, participants)
+
+
+def dump_comments(comments):
+    """
+    Dump comments data into Django database.
+    """
+    dump_data(Comment, comments)
+
+
+def dump_reactions(reactions):
+    """
+    Dump reactions data into Django database.
+    """
+    dump_data(Reaction, reactions)
--- a/meta_review/handler.py
+++ b/meta_review/handler.py
@@ -1,0 +1,485 @@
+from dateutil import parser
+import json
+import logging
+import os
+
+from django.utils import timezone
+
+from meta_review.dump_to_db import dump_all
+from meta_review.load_from_db import load_all
+from meta_review.models import Participant
+
+
+def parse_time(time):
+    """
+    Parse string to datetime.
+
+    :param time: a string represents time, e.g. 2018-05-09T11:19:26Z
+    :return: an offset-aware datetime object
+    """
+    if time is None:
+        return None
+
+    return parser.parse(time)
+
+
+class MetaReviewHandler:
+    """
+    Handle meta-review system.
+
+    This is the class responsible for scraping provided information (reviews,
+    reactions), processing them and dumping into Django database.
+    """
+
+    # coefficients of the scoring formula
+    BONUS_GIVE_POS = 0.05
+    BONUS_GIVE_NEG = 0.2
+
+    # point offset due to edited review after meta-review
+    LATE_EDIT_SCORE_OFFSET = 0.5
+
+    def __init__(self, content, date):
+        """
+        Construct a new ``MetaReviewHandler``.
+
+        :param content: Parsed JSON data
+        :param date: The update date
+        """
+        self.logger = logging.getLogger(__name__)
+        self.logger.info('this package is alive')
+
+        self.date = date
+
+        # save comments in memory
+        self.comments = dict()
+        for issue in content:
+            issue = issue['issue']
+            if not issue.get('pullRequest'):
+                continue
+            for comment in issue['pullRequest']['comments']:
+                if not comment.get('reactions') or not comment['reactions']:
+                    # there is no need to store comments which have
+                    # not yet been meta-reviewed
+                    continue
+                # parse time
+                comment['createdAt'] = parse_time(comment['createdAt'])
+                comment['lastEditedAt'] = parse_time(comment['lastEditedAt'])
+                self.comments[comment['id']] = comment
+
+        # save reactions in memory
+        self.reactions = dict()
+        for comment in self.comments.values():
+            for reaction_data in comment['reactions']:
+                # record receiver
+                reaction_data['receiver'] = {
+                    'login': comment['author']['login']
+                }
+                # record comment id
+                reaction_data['comment_id'] = comment['id']
+                # parse time
+                reaction_data['createdAt'] = parse_time(
+                    reaction_data['createdAt'])
+                self.reactions[reaction_data['id']] = reaction_data
+
+        # save participants in memory
+        self.participants = dict()
+        for comment in self.comments.values():
+            # get author of each comment
+            author = comment['author']
+
+            # skip if user does not exist
+            # this happens when account is deleted from GitHub
+            if author['login']:
+                self.participants[author['login']] = author
+
+            if not comment.get('reactions'):
+                continue
+            for reaction in comment['reactions']:
+                # get user of each reaction
+                user = reaction['user']
+
+                # skip if user does not exist
+                if user['login']:
+                    self.participants[user['login']] = user
+
+    def handle(self):
+        """
+        Scrape data, process and store in database.
+        """
+        is_first_deploy = self._is_first_deploy()
+
+        load_all(self.participants, self.comments, self.reactions)
+        dump_all(self.participants, self.comments, self.reactions)
+
+        self._update_scores()
+        self._update_weight_factors()
+
+        if is_first_deploy:
+            # dump first to make use of reverse query
+            dump_all(self.participants, self.comments, self.reactions)
+            self._recalculate_score()
+
+        self._apply_negative_scores()
+        self._update_time()
+
+        # dump first to make use of built-in sort method
+        dump_all(self.participants, self.comments, self.reactions)
+        self._update_rankings()
+
+        dump_all(self.participants, self.comments, self.reactions)
+
+        self.logger.info('Meta Review System finishes.')
+
+    def get_comments_modified_after_meta_review(self):
+        """
+        Generator for comments modified after meta-review.
+
+        Yield comments which are modified after they have been meta-reviewed.
+        """
+        for comment in self.comments.values():
+            last_edited_at = comment.last_edited_at
+            author = comment.author
+
+            # skip if author does not exist
+            if not author:
+                continue
+
+            comment_modified_after_meta_review = False
+            reactions = comment.reaction_set.all()
+
+            # get reactions createdTime
+            for reaction in reactions:
+                if last_edited_at and last_edited_at > reaction.created_at:
+                    comment_modified_after_meta_review = True
+                    break
+
+            if comment_modified_after_meta_review:
+                self.logger.info('%s updates review comment after it has been '
+                                 'meta-reviewed. Comment id: %s'
+                                 % (author.login, comment.id))
+                yield comment
+
+    def _apply_negative_scores(self):
+        """
+        Give point offset to participants who did improper behavior.
+
+        If an author updates his review comment after it has been
+        meta-reviewed, he will be slightly punished.
+        """
+        for participant in self.participants.values():
+            # give punished score back to the participant first,
+            # as offset will be recalculated later
+            participant.score += participant.offset
+            participant.modified_comments_after_meta_review = 0
+            participant.offset = 0
+
+        for comment in self.get_comments_modified_after_meta_review():
+            author = comment.author
+            self.logger.info('%s has improper behavior'
+                             '(edit comment after it has been '
+                             'meta-reviewed) on comment %s. %.2f '
+                             'point deducted.'
+                             % (author.login, comment.id,
+                                self.LATE_EDIT_SCORE_OFFSET))
+            author.score -= self.LATE_EDIT_SCORE_OFFSET
+            author.modified_comments_after_meta_review += 1
+            author.offset += self.LATE_EDIT_SCORE_OFFSET
+
+    def _update_time(self):
+        """
+        Update last_active_at attribute of each participant.
+
+        Note this does not accurately reflect the last time they were
+        active in the community.
+
+        First, it relies on the accuracy of issues.json fetched from
+        gh-board repo.
+
+        Second, this field should instead be interpreted as 'the last
+        time the participant had impact on the meta-review system'. This
+        is the last time among three things: the last time they created/edited
+        a comment, the last time they did a meta-review, the last time
+        their review received a meta-review.
+        """
+        self.logger.info('start updating last active time of all participants')
+
+        for participant in self.participants.values():
+            old_active_time = participant.last_active_at
+
+            # check last time they created/edited a comment
+            for comment in participant.comment_set.all():
+                if participant.last_active_at is None:
+                    participant.last_active_at = comment.created_at
+                if comment.created_at > participant.last_active_at:
+                    participant.last_active_at = comment.created_at
+                if (comment.last_edited_at and
+                        comment.last_edited_at > participant.last_active_at):
+                    participant.last_active_at = comment.last_edited_at
+
+            # check last time they did a meta-review
+            for reaction in participant.give.all():
+                if participant.last_active_at is None:
+                    participant.last_active_at = reaction.created_at
+                if reaction.created_at > participant.last_active_at:
+                    participant.last_active_at = reaction.created_at
+
+            # check last time they received a meta-review
+            for reaction in participant.receive.all():
+                if participant.last_active_at is None:
+                    participant.last_active_at = reaction.created_at
+                if reaction.created_at > participant.last_active_at:
+                    participant.last_active_at = reaction.created_at
+
+            if participant.last_active_at != old_active_time:
+                self.logger.debug('%s last active time changed from %s to %s'
+                                  % (participant.login, old_active_time,
+                                     participant.last_active_at))
+
+    @staticmethod
+    def _is_first_deploy():
+        """
+        Check whether it is the first deploy.
+
+        If the scoring system has not been run before, we need to calculate
+        draft score first, and then weight factor. Finally we use that
+        to recalculate score.
+
+        :return: a bool indicates whether this is the first deploy.
+        """
+        # This query set is empty means it is the first deploy.
+        return not Participant.objects.all()
+
+    def _recalculate_score(self):
+        """
+        Recalculate score of each participant.
+
+        Used only on first deploy. Based on the weight factor derived,
+        recalculate score for all participants and comments.
+        """
+        self.logger.info('recalculate score of each participant')
+        for comment in self.comments.values():
+            comment.clear_score()
+
+        for participant in self.participants.values():
+            participant.clear_score()
+
+        self._update_scores()
+        self._update_weight_factors()
+
+    def _update_scores(self):
+        """
+        Update score of each participant.
+
+        Calculate and update score of each participant using
+        the following formula:
+
+        Define:
+
+        P1 = total points (weighted) of THUMBS_UP a person gets for all
+             reviews he did.
+        P2 = total number of THUMBS_UP a person gives to other
+             people for their reviews.
+        N1 = total points (weighted) of THUMBS_DOWN a person gets for all
+             reviews he did.
+        N2 = total number of THUMBS_DOWN a person gives to other people for
+             their reviews.
+
+        Then final score, denote by S, is as follows:
+
+        S = P1 - N1 + BONUS_GIVE_POS * P2 + BONUS_GIVE_NEG * N2
+
+        where BONUS_GIVE_POS = 0.05, BONUS_GIVE_NEG = 0.2. One will get at
+        least 0.1 point for a positive reaction they received, so we want
+        BONUS_GIVE_POS to be smaller than that. BONUS_GIVE_NEG is larger
+        because people are reluctant to give negative reactions.
+        In all, bonus points (P2 and N2) aim to encourage people to do
+        meta-reviews, but we don't want them to dominate.
+
+        Also update score of each review comment.
+        """
+        self.logger.info('update scores of all participants')
+
+        for participant in self.participants.values():
+            # parameters to be used in the formula
+            p1, p2, n1, n2 = 0, 0, 0, 0
+
+            # number of positive/negative reactions received
+            pos_cnt, neg_cnt = 0, 0
+
+            time = participant.last_active_at
+            # get reactions received
+            if not time:
+                reactions_in = participant.receive.all()
+            else:
+                reactions_in = participant.receive.filter(created_at__gt=time)
+            for reaction in reactions_in:
+                # get weight factor of the reaction giver
+                weight_factor = reaction.giver.weight_factor
+                if reaction.content.find('THUMBS_UP') != -1:
+                    self.logger.debug('reaction received is %s, positive'
+                                      % reaction.content)
+                    p1 += weight_factor
+                    pos_cnt += 1
+                    # also update score of review comment
+                    reaction.review.pos += 1
+                    reaction.review.weighted_pos += weight_factor
+                    reaction.review.score += weight_factor
+                elif reaction.content.find('THUMBS_DOWN') != -1:
+                    self.logger.debug('reaction received is %s, negative'
+                                      % reaction.content)
+                    n1 += weight_factor
+                    neg_cnt += 1
+                    # also update score of review comment
+                    reaction.review.neg += 1
+                    reaction.review.weighted_neg += weight_factor
+                    reaction.review.score -= weight_factor
+                else:
+                    self.logger.debug('reaction received is %s, ignore'
+                                      % reaction.content)
+
+            # get reactions give away
+            if not time:
+                reactions_out = participant.give.all()
+            else:
+                reactions_out = participant.give.filter(created_at__gt=time)
+            for reaction in reactions_out:
+                if reaction.content.find('THUMBS_UP') != -1:
+                    self.logger.debug('reaction give away is %s, positive'
+                                      % reaction.content)
+                    p2 += 1
+                elif reaction.content.find('THUMBS_DOWN') != -1:
+                    self.logger.debug('reaction give away is %s, negative'
+                                      % reaction.content)
+                    n2 += 1
+                else:
+                    self.logger.debug('reaction give away is %s, ignore'
+                                      % reaction.content)
+
+            # update information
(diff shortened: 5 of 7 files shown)


li-boxuan:
This PR contains two commits. The first commit is on https://github.com/coala/community/pull/143. When that PR is merged, I would remove the first commit from this PR.

.travis.yml
```diff
@@ -22,9 +22,9 @@ before_install:
   - npm i -g jshint
 
 script:
+  - coala --non-interactive -V
   - ./.ci/build.sh
   - rm -rf private/
-  - coala --non-interactive -V
```
blazeu: Any reason for the reordering, or is it still WIP ?

li-boxuan: It is ready for review. It has passed Travis (it is failing due to API limit, but it has passed it a few times), but failed on Netlify (I am still investigating that).

The reordering is because of this issue https://github.com/coala/coala-bears/issues/2522. `python manage.py run_meta_review_system` will generate a large file (20MB), which makes `coala` gets stuck. The ordering eschews the problem. Also, I think there is justification to do so: we shouldn't run coala on files generated by `.ci/build.sh` IMO.

blazeu: I think it's worth adding a comment with a link to the issue, or just explain it in the commit.

li-boxuan:
Netlify deploy is failing because of the time limit. It takes Travis 9 minutes while it takes Netlify more than 20 minutes, due to the slowness of CPU speed.

See https://app.netlify.com/sites/community-li-boxuan/deploys/5b19f9430733d559b900ed93 for the verbose log.

The bottleneck is this function: https://github.com/coala/community/pull/143/commits/2fc319d8a3b018343fc9e8b8c7d4f2adde1efe9a#diff-fa5d2660d275d02c59979fe967702a7bR203, not sure how to optimize that. Multithreading or multiprocessing? Need suggestion.

meta_review/data.py
```diff
@@ -0,0 +1,667 @@
...
+
+                # skip if user not exist
+                if author['login']:
+                    self.participants[user['login']] = user
+
+    def handle(self):
+        """
+        Scrape data, process and store in database
+        """
+        print('load participants to memory')
```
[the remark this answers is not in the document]
li-boxuan: Hmm, you could see I am using logging in most places. I use `print` here to see Netlify deploy process. The background is Netlify deploy process is much slower than Travis when running this script. If the deploy time exceeds 15 mins, Netlify deploy will fail, and we have no way to see the log if using logging.

meta_review/data.py
```diff
@@ -0,0 +1,666 @@
...
+                reaction['createdAt'] = parse_time(reaction['createdAt'])
+                self.reactions[reaction['id']] = reaction
+
+        # save participants in memory
+        self.participants = dict()
+        for key, comment in self.comments.items():
+            # get author of each comment
+            author = comment['author']
+
+            # skip if user not exist
```
andrewda:
```python
# skip if user does not exist
```

meta_review/data.py
```diff
@@ -0,0 +1,666 @@
...
+            if author['login']:
+                self.participants[author['login']] = author
+
+            if not comment.get('reactions'):
+                continue
+            for reaction in comment['reactions']:
+                # get user of each reaction
+                user = reaction['user']
+
+                # skip if user not exist
```
andrewda: same as above

meta_review/data.py
```diff
@@ -0,0 +1,666 @@
...
+        issues_url = 'https://' + org_name + '-gh-board.netlify.com/issues.json'
+        content = requests.get(issues_url, timeout=10)
+
+    try:
+        parsed_json = content.json()
+    except json.JSONDecodeError:
+        logger.error('JSON decode error')
+
+    handler = MetaReviewHandler(parsed_json['issues'], timezone.now())
+    handler.handle()
```
andrewda: Might want to lump these two lines into the `try` above so that you don't get an `UnboundLocalError` if the above fails (see: Travis failure).

meta_review/data.py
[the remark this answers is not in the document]
li-boxuan: I'll remove them for now, as deploy runtime is not an issue anymore.

meta_review/data.py
```diff
@@ -0,0 +1,657 @@
...
+        except Exception as ex:
+            logger.error('Get issues from %s failed: %s'
+                         % (issues_url, ex))
+
+    try:
+        parsed_json = content.json()
+        handler = MetaReviewHandler(parsed_json['issues'], timezone.now())
+        handler.handle()
+    except json.JSONDecodeError as ex:
+        logger.error('JSON decode error: %s' % ex)
```
hemangsk: Consider breaking this down to three files, one for handlers, one for all the update functions and one for all the loader functions ?

.travis.yml
```diff
@@ -22,9 +22,11 @@ before_install:
   - npm i -g jshint
 
 script:
+  # Run coala before build.sh due to the issue
```
jayvdb: do you know which files are causing it?
can you exclude those from the keyword search?
that would be better.

gci/management/commands/fetch_deployed_data.py
```diff
@@ -13,12 +13,15 @@ class Command(BaseCommand):
...
-        parser.add_argument('filenames', nargs='+', type=str)
+        parser.add_argument('filenames', nargs='?', type=str)
+        parser.add_argument('allow_failure', nargs='?', type=bool,
+                            default=False)
 
     def handle(self, *args, **options):
         logger = logging.getLogger(__name__)
         output_dir = options.get('output_dir')
-        filenames = options.get('filenames')
+        filenames = options.get('filenames').split(' ')
```
jayvdb: Im doing https://github.com/coala/community/issues/132 ..

But .. this is not desirable

It is better to let them be separate args.

If you want to use `allow_failure`, call it on the cli like `--allow_failure`

meta_review/handler.py
```diff
@@ -0,0 +1,440 @@
...
+
+            if need_punishment:
+                self.logger.info('%s updates review comment after it has been'
+                                 'meta-reviewed. 0.5 point deducted for'
+                                 'punishment. comment id: %s'
+                                 % (author.login, comment.id))
+                author.punishment += 0.5
+                author.score -= 0.5
+
+    def __update_time(self):
```
jayvdb: why `__` prefix?

meta_review/handler.py
```diff
@@ -0,0 +1,440 @@
...
+                participant.weight_factor *= 0.9
+                participant.weight_factor += 0.1
+
+
+def handle():
+    # load data from gh-board repo
+    org_name = get_org_name()
+
+    # URL to grab all issues from
+    issues_url = 'http://' + org_name + '.github.io/gh-board/issues.json'
```
jayvdb: The fetch operation should be in a separate management command, so that people can fetch once and then load many times while attempting to debug problems.

This could probably be an enhancement to `fetch_deployed_data` (and `community.git`) to fetch deployed data from a different repo.

meta_review/handler.py
```diff
@@ -0,0 +1,440 @@
...
+
+    logger = logging.getLogger(__name__)
+
+    try:
+        content = requests.get(issues_url, timeout=10)
+        content.raise_for_status()
+    except Exception as ex:
+        logger.warning('Get issues from %s failed: %s Try backup url'
+                       % (issues_url, ex))
+        issues_url = 'https://' + org_name + '-gh-board.netlify.com/issues.json'
```
jayvdb: If this fallback mechanism is moving to `community.git`, it should work correctly for the `community` repo, as that netlify is called `coala-community`, so the naming convention is the same.

.ci/build.sh
```diff
@@ -6,16 +6,36 @@ mkdir private _site public
...
-  python manage.py fetch_deployed_data _site $EXPORTED_DATA
+  python manage.py fetch_deployed_data _site "$EXPORTED_DATA"
 fi
 
 python manage.py migrate
 python manage.py import_contributors_data
 python manage.py import_openhub_data
+
+# fetch deployed meta_review data
+python manage.py fetch_deployed_data _site "$META_REVIEW_DATA" True
```
jayvdb: this goes before `python manage.py migrate`

.ci/build.sh
```diff
@@ -6,16 +6,36 @@ mkdir private _site public
...
+if [[ -f "$META_REVIEW_DATA" ]]; then
+  echo "File $META_REVIEW_DATA exists."
+  # Load meta_review data from json
+  python manage.py loaddata "$META_REVIEW_DATA"
+else
+   echo "File $META_REVIEW_DATA does not exist."
+fi
+
+# Run meta review system
+python manage.py run_meta_review_system
```
jayvdb: split into two commands

1. for fetching external data, like `issues.json`, which goes above `python manage.py migrate`
2. processing the external data, which goes after `python manage.py migrate`, and then the dump data below it

meta_review/handler.py
li-boxuan: This should be a private method IMO.

.ci/build.sh
li-boxuan: I am considering making use of `fetch_deployed_data` to retrieve external data (issues.json), and save it in static directory. The extra benefit is that in case `gh-board` repo breaks down, we still have raw data when debugging.

.travis.yml
li-boxuan: This is not a problem anymore. I dumped `meta_review.json` to `static` folder, which should be in `_site` folder. It is fixed now.

meta_review/handler.py
```diff
@@ -0,0 +1,412 @@
...
+
+class MetaReviewHandler:
+    """
+    This is the class responsible for scraping provided information (reviews,
+    reactions), processing them and dumping into Django database
+    """
+
+    def __init__(self, content, date):
+        """
+        Constructs a new ``MetaReviewHandler``
```
jayvdb: run pydocstyle on your code.  It will tell you to add a period at the end here.

Ideally get the relevant bear working on this repo when you can.

meta_review/handler.py
```diff
@@ -0,0 +1,412 @@
...
+                self.participants[author['login']] = author
+
+            if not comment.get('reactions'):
+                continue
+            for reaction in comment['reactions']:
+                # get user of each reaction
+                user = reaction['user']
+
+                # skip if user does not exist
+                if author['login']:
```
jayvdb: should this be `user['login']` ?

meta_review/handler.py
jayvdb: private from whom?

It is an entirely open system, using public data.

What is the benefit of using `__` instead of weak private methods?
what is the risk model?

Anyone wanting to run this in a non-static system isnt going to have a more secure/stable system because of this, as some other part of code could access `.comments`, `.participants`, etc

OTOH, using `__` makes it harder to develop and test due to the mangling. Note that even the mangling doesnt stop access - just makes it a little bit harder.

meta_review/handler.py
```diff
@@ -0,0 +1,412 @@
...
+        dump_all(self.participants, self.comments, self.reactions)
+
+        self.logger.info('Meta Review System finishes.')
+
+    def __check_comment_update(self):
+        """
+        If reviewer updates their review comment after it has
+        been meta-reviewed, they will be slightly punished.
+        """
+        for key, comment in self.comments.items():
```
jayvdb: "check_comment_update" isnt a good name for a method which modifies state based on complex logic.

The logic for determining who modified their review comments after meta-review should be separate from punishment.

It is very likely that other apps, especially the gamification app, might be interested in using the list of who is editing their comments.

Also the meta-review system might use it in multiple ways.  e.g. in the future, we might want to count how many times they are doing it, and log warnings for people who are regularly doing it, as the punishment isnt working.

meta_review/handler.py
```diff
@@ -0,0 +1,412 @@
...
+        positive reaction they received, so we want c1 be smaller than that.
+        c2 is larger because people are reluctant to give negative reactions.
+        In all, bonus points (P2 and N2) aim to encourage people to do
+        meta-reviews, but we don't want them to dominate.
+
+        Also update score of each review comment.
+        """
+        self.logger.info('update scores of all participants')
+        # coefficients of the formula
+        c1, c2 = 0.05, 0.2
```
jayvdb: store as CONSTANTs at top of module or CONSTANTs at the class level.

li-boxuan:
> It is a bit strange that incorrectusername (0 positive interactions received) appears before vijeth-aradhya (3 positive interactions received).

This is because the scoring formula I am using. People get reward points when they do meta-reviews. `incorrectusername` did 8 meta-reviews while `vijeth-aradhya` only did 2. Welcome any suggestion on the formula, especially coefficients.

> you can detect that this is the first run, and do __update_weight_factors twice.

Although this is not the reason for the phenomenon you observed, this suggestion sounds great. Will do that.

meta_review/handler.py
li-boxuan: Yeah, a bug here. This would lead to an exception if the user of reaction does not exist. Either GitHub already handled this, or we just didn't encounter it.

.ci/build.sh
li-boxuan: Agree. However, saving `issues.json` on the deployed website has other benefits:
1. Other apps could reuse that file. There is no need to fetch that file again.
2. The up-to-date `issues.json` on gh-board repo may be different from what the scoring system used during deployment. 'Caching' `issues.json` makes debugging easier.

meta_review/handler.py
```diff
@@ -0,0 +1,499 @@
...
+class MetaReviewHandler:
+    """
+    Handle meta-review system.
+
+    This is the class responsible for scraping provided information (reviews,
+    reactions), processing them and dumping into Django database.
+    """
+
+    # coefficients of the scoring formula
+    C1, C2 = 0.05, 0.2
```
jayvdb: better names?

li-boxuan: I'm pretty bad at naming. What about `BONUS_GIVE_POS` and `BONUS_GIVE_NEG`?

meta_review/handler.py
```diff
@@ -0,0 +1,499 @@
...
+        self.logger.info('Meta Review System finishes.')
+
+    def _punish_improper_behavior(self):
+        """
+        Give punishment points to participants who did improper behavior.
+
+        If an author updates his review comment after it has been
+        meta-reviewed, he will be slightly punished.
+        """
+        for key, comment in self.comments.items():
```
jayvdb: This is looking better, but it still mixes analysis algorithms with business logic.

Create an extra public method which returns a generator of the modified-after-meta-review comments.  Then it is able to be documented as a dataset, and the documentation will also help it be reused because other people know about this list of people.

You could even have another method which converts the list of 'modified-after-meta-review comments' into a dict of `{author: count_of_edited_comments_after_meta_review}` (or adds `author.count_of_edited_comments_after_meta_review`?), and then here you can very easily log and update those authors.

meta_review/handler.py
```diff
@@ -0,0 +1,499 @@
...
+            if self._check_comment_update(comment):
+                author = comment.author
+                self.logger.info('punish %s for their improper behavior '
```
jayvdb: be specific : for editing comment after meta-review

we will add more negative score types.

meta_review/handler.py
```diff
@@ -0,0 +1,499 @@
...
+        # clear score of all comments and participants
+        for key, comment in self.comments.items():
+            comment.pos = 0
+            comment.weighted_pos = 0
+            comment.neg = 0
+            comment.weighted_neg = 0
+            comment.score = 0
+
+        for key, participant in self.participants.items():
+            participant.pos_in = 0
```
jayvdb: create a `Participant` method which does this.

meta_review/handler.py
```diff
@@ -0,0 +1,499 @@
...
+        """
+        Recalculate score of each participant.
+
+        Used only on first deploy. Based on the weight factor derived,
+        recalculate score for all participants and comments.
+        """
+        self.logger.info('recalculate score of each participant')
+        # clear score of all comments and participants
+        for key, comment in self.comments.items():
+            comment.pos = 0
```
jayvdb: create a `Comment` method which does this.

meta_review/handler.py
```diff
@@ -0,0 +1,499 @@
...
+
+    def _recalculate_score(self):
+        """
+        Recalculate score of each participant.
+
+        Used only on first deploy. Based on the weight factor derived,
+        recalculate score for all participants and comments.
+        """
+        self.logger.info('recalculate score of each participant')
+        # clear score of all comments and participants
```
jayvdb: if you create methods on Comment and Participant, the code is self-documenting and this comment is redundant.

meta_review/handler.py
```diff
@@ -0,0 +1,499 @@
...
+        for key, participant in self.participants.items():
+            participant.pos_in = 0
+            participant.weighted_pos_in = 0
+            participant.neg_in = 0
+            participant.weighted_neg_in = 0
+            participant.pos_out = 0
+            participant.neg_out = 0
+            participant.score = 0
+
+        # recalculate score and weight factor
```
jayvdb: this is a redundant comment, which is discouraged in Python in favour of readable code, and the next two lines are very readable.

meta_review/load_from_db.py
```diff
@@ -0,0 +1,190 @@
+import logging
+
+from meta_review.models import Participant, Comment, Reaction
+
+
+def load_all(participants, comments, reactions):
+    '""Load all meta-review related data to memory.""'
```
jayvdb: `'"" ... ""'` is invalid.

Our own docs bear has a bug (and someone is fixing it).  There is a `pydocstyle` bear, which should be used instead.

Or to be compliant with both bears, simply avoid the single line syntax, which is then consistent with the other docstrings which require more explanation.
```
"""
Foo.
"""

File: meta_review/dump_to_db.py [hunk 26]
```diff
@@ -0,0 +1,61 @@
+import logging
+
+from meta_review.models import Participant, Comment, Reaction
+
+
+def dump_all(participants, comments, reactions):
+    '""Dump all meta-review related data into database.""'
```
jayvdb: see note at bottom re `'""`.

meta_review/handler.py
```diff
@@ -0,0 +1,499 @@
...
+            # parameters to be used in the formula
+            p1, p2, n1, n2 = 0, 0, 0, 0
+
+            # number of positive/negative reactions received
+            pos_cnt, neg_cnt = 0, 0
+
+            # get reactions received
+            reactions_in = participant.receive.all()
+            for reaction in reactions_in:
+                # skip old reactions since they were counted before
```
jayvdb: create a method which is a generator of only new reactions not seen before.

That separates the logic regarding sync with the logic of processing items.

meta_review/handler.py
```diff
@@ -0,0 +1,499 @@
...
+                    reaction.review.weighted_neg += weight_factor
+                    reaction.review.score -= weight_factor
+                else:
+                    self.logger.debug('reaction received is %s, ignore'
+                                      % reaction.content)
+
+            # get reactions give away
+            reactions_out = participant.give.all()
+            for reaction in reactions_out:
+                # skip old reactions since they were counted before
```
jayvdb: re-use generator here.

meta_review/dump_to_db.py
```diff
@@ -0,0 +1,61 @@
...
+        logger.error('Something went wrong saving participants: %s' % ex)
+
+
+def dump_comments(comments):
+    '""Dump comments data into Django database.""'
+    logger = logging.getLogger(__name__)
+    logger.info('dump review comments data into database')
+    comments_list = []
+    for key, comment in comments.items():
+        comments_list.append(comment)
```
jayvdb: do we need to store comments which have not been meta-reviewed?
or have they already been excluded somewhere else?

It would be helpful to avoid storing them, to reduce spreading content which is of low value.

Think GDPR ... what happens if we're asked to remove all information about a person -- the git history would need to be sanitised.  Best to only copy data that we can justify keeping a copy of.

meta_review/handler.py
```diff
@@ -0,0 +1,499 @@
...
+
+        # dump first to make use of built-in sort method
+        dump_all(self.participants, self.comments, self.reactions)
+        self._update_rankings()
+
+        dump_all(self.participants, self.comments, self.reactions)
+
+        self.logger.info('Meta Review System finishes.')
+
+    def _punish_improper_behavior(self):
```
jayvdb: punish.. -> apply_negative_scores

.ci/build.sh
jayvdb: Copies cause problems.  Think GDPR.

jayvdb:
webservices is broken atm.  see logs
Maybe create a workaround for that ;-)

li-boxuan:
Hmmm, I don't quite understand the log. Why is Travis running `python manage.py test`, which isn't on the `.ci/build.sh`?

andrewda:
@li-boxuan `python manage.py test` is in the build script:

https://github.com/coala/community/blob/d8e760bf8abd555ef643652bdcdf652fda5575ee/.ci/build.sh#L18

andrewda:
@li-boxuan Travis runs the build on the result of a merge between your branch and the branch it's merging into if possible. If you update your branch to be in-sync with `master`, you'll see the `python manage.py test`.

https://github.com/coala/community/blame/d8e760bf8abd555ef643652bdcdf652fda5575ee/.ci/build.sh#L18

jayvdb:
> webservices is broken atm. see logs
> Maybe create a workaround for that ;-)

This PR is marked as blocked.
Fix the blocker before updating this PR again.
https://github.com/coala/community/issues/154

meta_review/dump_to_db.py
jayvdb: Addressed around https://github.com/coala/community/pull/143/files#diff-745f0342d354672f4a930f0a8bc86590R60

data/management/commands/fetch_external_data.py
```diff
@@ -0,0 +1,58 @@
...
+    def handle(self, *args, **options):
+        logger = logging.getLogger(__name__)
+        output_dir = options.get('output_dir')
+        repo_name = options.get('repo_name')
+        filenames = options.get('filenames')
+        allow_failure = options.get('allow_failure')
+        org_name = get_org_name()
+
+        for filename in filenames:
+            url = ('http://' + org_name + '.github.io/'
```
jayvdb: please enhance `community.git` , so that you are re-using logic of `get_deploy_url`/`get_upstream_deploy_url`.  Probably this can be done by adding args to `get_upstream_deploy_url`, or maybe a new function is needed.

data/management/commands/fetch_external_data.py
```diff
@@ -0,0 +1,58 @@
...
+                   + repo_name + '/' + filename)
+            logger.info('Get external data from %s' % url)
+            try:
+                r = requests.get(url, timeout=10)
+                r.raise_for_status()
+            except Exception as ex:
+                logger.warning('Get external data from %s failed: %s '
+                               'Try backup url' % (url, ex))
+                url = ('https://' + org_name + '-'
```
jayvdb: again, please build this in `community.git`.
already requested once at https://github.com/coala/community/pull/143#discussion_r194249685

netlify is just another potential host.

We also need `community.git` to support 'gitlab pages'.
(e.g. for orgs which are GitLab only)
And it would be great if orgs can choose which hoster they are using, and switch between then.  e.g. my real job, NGO https://ciptamedia.org/ is @cipta-media and https://gitlab.com/ciptamedia and also on Netlify , and it is built so that it could be deployed on any of them, so that changes of Terms of Use or a hoster charging for the service are not a gun-at-our-head; instead it is a slightly annoying retest and DNS update to discard a host which is no longer suitable. (I guess we could also add BitBucket for extra redundancy, and possibly others .. ;-)

I'm still not seeing the benefit of allowing two potential fetch locations for the same file, and `allow_failure`.
This is hiding failure, and the assumption that the backup is OK is not very sane IMO.  Likely the primary gets broken, and nobody fixes it because there is a backup (or a PR gets ignored in the queue), and then the backup breaks, and then the primary gets fixed but the backup stays broken, and then ... ;-)

The problems with the data not being retained in one of those hosters is a problem with the build system of gh-board, and one of the solutions is incredibly easy : https://github.com/coala/gh-board/issues/52  mentioned previously at https://github.com/coala/community/pull/143#discussion_r194706920 and at https://github.com/coala/community/pull/143#discussion_r194249921

If we remove this dual host thing, we could easily add the `fetch_external_data.py` functionality into the `fetch_deployed_data.py` command by adding an optional 'repo' argument.

failure is our friend.  exceptions with full backtrace are useful to get operational quickly.

meta_review/handler.py
```diff
@@ -0,0 +1,488 @@
...
+
+    This is the class responsible for scraping provided information (reviews,
+    reactions), processing them and dumping into Django database.
+    """
+
+    # coefficients of the scoring formula
+    BONUS_GIVE_POS, BONUS_GIVE_NEG = 0.05, 0.2
+
+    # punishment point due to edited review after meta-review
+    PUNISHMENT_EDIT = 0.5
```
jayvdb: use a neutral term.  no words like punishment.

`LATE_EDIT_SCORE_OFFSET`

li-boxuan:
Blocked by https://github.com/coala/community/issues/161

community/git.py
```diff
@@ -179,3 +179,23 @@ def get_upstream_deploy_url():
...
+
+def get_external_url(org_name, repo_name, file_name):
+    """
+    Obtain external url
+    """
+    # Use environment variable to switch between sources
+    external_site = os.environ.get('EXTERNAL_SITE')
+    if external_site is None:
+        # use GitHub as default
+        url = ('http://' + org_name + '.github.io/'
```
jayvdb: This logic is `get_upstream_deploy_url()` above.

And if you use that, IIRC, you should not need `EXTERNAL_SITE` because it lets forks have their own website, and we can enhance it if it isnt working for you.

But Netlify support could be added, now.

GitLab support should be done when the rest of the module supports GitLab as a remote. (very separate issue)

And if you add arg `path` to `get_upstream_deploy_url()` , it is probably exactly what you need.

data/management/commands/fetch_deployed_data.py
```diff
@@ -14,12 +14,14 @@ class Command(BaseCommand):
...
         parser.add_argument('output_dir', nargs='?', type=str)
         parser.add_argument('filenames', nargs='+', type=str)
+        parser.add_argument('--allow_failure', action='store_true',
+                            default=False, help='Don\'t raise exceptions.')
 
     def handle(self, *args, **options):
         logger = logging.getLogger(__name__)
         output_dir = options.get('output_dir')
         filenames = options.get('filenames')
-
```
jayvdb: readd the blank line below the new option

data/management/commands/fetch_deployed_data.py
```diff
@@ -34,9 +36,18 @@ def handle(self, *args, **options):
             except Exception:
                 if upstream_deploy_url:
                     r = requests.get(upstream_deploy_url + '/' + filename)
-                    r.raise_for_status()
+                    try:
+                        r.raise_for_status()
+                    except Exception:
```
jayvdb: you are not using the exception, soo ..

`if not allow_failure: r.raise_for_status()`

??

But I need to see this justification for allowing the failure....

meta_review/migrations/0002_auto_20180707_0305.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('meta_review', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RenameField(
```
jayvdb: this is should be a separate commit; and probably better to be a separate PR so it goes in straight away and your final PR doesnt include this.

li-boxuan: Separate PR created: https://github.com/coala/community/pull/166

community/git.py
li-boxuan: For this PR, Netlify is not needed. I can remove `get_external_url`, reuse and add `owner`, `path` options to `get_upstream_deploy_url`. One problem I am wondering is that would break the semantic meanging of `upstream_deploy_url`.

li-boxuan:
Will remove https://github.com/coala/community/pull/143/commits/1f430da573a7e8e5e87a1fa00baf5a54303221f0 when https://github.com/coala/community/pull/166 is merged.

li-boxuan:
Will remove https://github.com/coala/community/pull/143/commits/1f430da573a7e8e5e87a1fa00baf5a54303221f0 when https://github.com/coala/community/pull/166 is merged.

data/management/commands/fetch_external_data.py
```diff
@@ -0,0 +1,41 @@
...
+
+from community.git import get_deploy_url
+
+
+class Command(BaseCommand):
+    help = 'Fetch external data'
+
+    def add_arguments(self, parser):
+        parser.add_argument('output_dir', nargs='?', type=str)
+        parser.add_argument('repo_name', nargs='?', type=str)
```
jayvdb: add this to fetch_deployed_data , and then we dont need a second command ?

meta_review/handler.py
```diff
@@ -0,0 +1,489 @@
...
+        [0.28, 0.172, 1.0]
+
+        Anyone who gets negative marks from previous run will have weight
+        factor of 0.
+
+        To conclude, the weight factor is a float number ranging from 0 to 1.
+        """
+        max_score = 1.0
+        # find max score
+        for key, participant in self.participants.items():
```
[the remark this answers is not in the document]
li-boxuan: I ran `pyflakes`, but it doesn't report any warning/error.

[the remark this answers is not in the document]
jayvdb: Ah yes ... the loop vars problem in pyflakes.

but in this case, it means that `.values()` should be used.

meta_review/dump_to_db.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def dump_participants(participants):
+    """
+    Dump participants data into Django database.
+    """
+    logger = logging.getLogger(__name__)
+    logger.info('dump participants data into database')
+    participants_list = []
+    for key, participant in participants.items():
```
jayvdb: use `.values()` if you dont need `key`

meta_review/dump_to_db.py
```diff
@@ -0,0 +1,69 @@
...
+
+def dump_comments(comments):
+    """
+    Dump comments data into Django database.
+    """
+    logger = logging.getLogger(__name__)
+    logger.info('dump review comments data into database')
+    comments_list = []
+    for key, comment in comments.items():
+        comments_list.append(comment)
```
jayvdb: use `.values()` if you dont need `key`

meta_review/dump_to_db.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def dump_reactions(reactions):
+    """
+    Dump reactions data into Django database.
+    """
+    logger = logging.getLogger(__name__)
+    logger.info('dump reactions data into database')
+    reactions_list = []
+    for key, reaction in reactions.items():
```
jayvdb: use `.values()` if you dont need `key`

meta_review/load_from_db.py
```diff
@@ -0,0 +1,192 @@
...
+    # load all participants again (old + new)
+    all_participants = Participant.objects.all()
+
+    for p in all_participants:
+        if participants.get(p.login):
+            # participants recently active
+            participant = participants[p.login]
+            p.name = participant['name']
+            # save into memory
+            participants[p.login] = p
```
jayvdb: `participants[p.login] = p` is common to both branches, so could be after the `if`

.ci/build.sh
```diff
@@ -6,6 +6,11 @@ mkdir private _site public
 
 EXPORTED_DATA="static/tasks.yaml static/instances.yaml"
 
+EXTERNAL_REPO="gh-board"
```
jayvdb: We dont need this variable.  It is only used once.

And we will have other external repos in the future.

.ci/build.sh
```diff
@@ -6,6 +6,11 @@ mkdir private _site public
...
+EXTERNAL_DATA="issues.json"
```
jayvdb: `ISSUES_JSON` ?

not sure this is needed

.ci/build.sh
```diff
@@ -14,9 +19,34 @@ else
   python manage.py fetch_deployed_data _site $EXPORTED_DATA
 fi
 
+# fetch deployed meta_review data
+python manage.py fetch_deployed_data --allow_failure \
```
jayvdb: You could add `static/meta_review.json` to `EXPORTED_DATA` , and *always* run `python manage.py fetch_deployed_data --allow_failure _site $EXPORTED_DATA`  *before* the `GCI_TOKEN` token command above.

(We didnt have `--allow_failure` / `--allow-failure` , so it needed the `if` )

data/management/commands/fetch_deployed_data.py
```diff
@@ -14,13 +14,19 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument('output_dir', nargs='?', type=str)
         parser.add_argument('filenames', nargs='+', type=str)
+        parser.add_argument('--repo_name', action='store',
+                            dest='repo_name', type=str)
+        parser.add_argument('--allow_failure', action='store_true',
```
jayvdb: this should be `--allow-failure` .

I should have done that with `output_dir`, but it is a nargs so the name isnt so critical.

data/management/commands/fetch_deployed_data.py
```diff
@@ -14,13 +14,19 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument('output_dir', nargs='?', type=str)
         parser.add_argument('filenames', nargs='+', type=str)
+        parser.add_argument('--repo_name', action='store',
```
jayvdb: `--repo-name`

.ci/build.sh
li-boxuan: `ISSUES_JSON` makes sense IMO, suppose we cut the large file into smaller files `issues_1.json`, `issues_2.json`, ... in the future

.ci/build.sh
```diff
@@ -4,19 +4,43 @@ set -e -x
 
 mkdir private _site public
 
-EXPORTED_DATA="static/tasks.yaml static/instances.yaml"
+META_REVIEW_DATA="meta_review.json"
+
+EXPORTED_DATA="static/tasks.yaml static/instances.yaml static/$META_REVIEW_DATA"
```
jayvdb: 81 chars

meta_review/dump_to_db.py
```diff
@@ -0,0 +1,69 @@
...
+    dump_reactions(reactions)
+
+
+def dump_participants(participants):
+    """
+    Dump participants data into Django database.
+    """
+    logger = logging.getLogger(__name__)
+    logger.info('dump participants data into database')
+    participants_list = []
```
jayvdb: `participants_list = participants.values()` ?

or ...

meta_review/dump_to_db.py
```diff
@@ -0,0 +1,69 @@
...
+    for participant in participants.values():
+        participants_list.append(participant)
+    try:
+        # there's no way to do bulk update, so we delete and then create
+        # bulk delete all participants
+        Participant.objects.all().delete()
+        # bulk create all participants
+        Participant.objects.bulk_create(participants_list)
```
jayvdb: `Participant.objects.bulk_create(participants.values())`

meta_review/dump_to_db.py
```diff
@@ -0,0 +1,69 @@
...
+        logger.error('Something went wrong saving participants: %s' % ex)
+
+
+def dump_comments(comments):
+    """
+    Dump comments data into Django database.
+    """
+    logger = logging.getLogger(__name__)
+    logger.info('dump review comments data into database')
+    comments_list = []
```
jayvdb: as above ?

.ci/build.sh
li-boxuan: nope, actually 80 chars

meta_review/dump_to_db.py
```diff
@@ -0,0 +1,69 @@
...
+        # there's no way to do bulk update, so we delete and then create
+        # bulk delete all comments
+        Comment.objects.all().delete()
+        # bulk create all comments
+        Comment.objects.bulk_create(comments_list)
+    except Exception as ex:
+        logger.error('Something went wrong saving comments: %s' % ex)
+
+
+def dump_reactions(reactions):
```
jayvdb: These three functions look identical ... could be done in a helper function
```py
def dump_data(cls, data):
    logger = logging.getLogger(__name__)
    logger.info('dump %s data into database' % cls.__name__.lower())
    ...
```

jayvdb:
ack https://github.com/coala/community/pull/143/commits/78e1e3a542a1654982f55cde4e08bbf97ec81c3c

jayvdb:
@gitmate-bot ff

Issue #150, opened 2018-06-12
Create tests for openhub app


sks444:

Fenn-CS:
@sks444  are you still working on this? @jayvdb  can I pick this up?

jayvdb:
@Fenn-CS , ya, we have tests, and this is 99% finished, but there are problems ... and we've failed many times to figure them out.  Today we disabled the OpenHub app, until we have more time to work them out.

If you are interested in OpenHub, and know a bit of Node, I could assign https://github.com/coala/gci-leaders/issues/166 to you.  I was about to schedule it to be given to GCI students, but they have tasks, so you could take it up.

Fenn-CS:
@jayvdb I see you have assigned this to @sks444  is that is a mistake? If not then you can assign it to me. Would work on it.

jayvdb:
@Fenn-CS , it is no mistake.  He has been working on this for a very long problem, and it is a hard problem.

Please read my comment again, and reply regarding coala/gci-leaders#166

Fenn-CS:
@jayvdb ok then I have some knowledge of Node! You can assign that to me. Meanwhile I have worked with tests for a while in pytest and django @sks444 can document the challenges that are faced and tag me to so we debug together.

jayvdb:
@Fenn-CS , you are now assigned to https://github.com/coala/gci-leaders/issues/166 .  No more comments on this issue please.

PR #152, opened 2018-06-13
openhub/data.py: Convert data dict to list

sks444:
This checks if type of 'data' is not list then
convert it into list, as 'data1' is a list
and to concatenate 'data' into 'data1', 'data'
must be a list.

Closes https://github.com/coala/community/issues/151

PR #153, opened 2018-06-13
openhub/tests/: Create tests for openhub app

sks444:
This adds tests for openhub app models, views
and commands.

Closes https://github.com/coala/community/issues/150


openhub/data.py
```diff
@@ -51,5 +51,7 @@ def get_data(for_what):
             except Exception as ex:
                 logger.error(ex)
                 break
+        if type(data) is not list:
```
sks444: This is a bug which is being fixed in https://github.com/coala/community/pull/152

I will remove it from here when that is merged.

sks444:
Not sure on why these tests are not being executed by `python manage.py test` , although its being executed by `python manage.py test openhub.tests` command.

openhub/tests/commands/test_import_affiliated_committers_data.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class ImportAffiliatedCommitterDataTest(TestCase):
+
+    @classmethod
+    def setUpTestData(cls):
+        call_command('import_affiliated_committers_data')
+
+    def test_command_import_affiliated_committers_data(self):
+        a_committers = AffiliatedCommitter.objects.all()
```
li-boxuan:
```python
committer = AffiliatedCommitter.objects.get(name='John Vandenberg')
```
This looks better to me, and is consistent with what you do in `test_import_organization_data.py`

openhub/tests/commands/test_import_organization_data.py
```diff
@@ -0,0 +1,15 @@
...
+
+class ImportOrganizationDataTest(TestCase):
+
+    @classmethod
+    def setUpTestData(cls):
+        call_command('import_organization_data')
+
+    def test_command_import_organization_data(self):
+        org = Organization.objects.get(name='coala')
+        self.assertIn('coala', org.name)
```
li-boxuan: Could be 
```python
self.assertIsNotNone(org)
```

openhub/tests/models/test_affiliated_committer_model.py
```diff
@@ -0,0 +1,90 @@
...
+            date='Jun 2018')
+        AffiliatedCommitter.objects.create(
+            name='Test User',
+            level=2,
+            org='coala',
+            most_commits=most_commits,
+            most_recent_commit=most_recent_commit)
+
+    def test_field_label(self):
+        affiliated_committer = AffiliatedCommitter.objects.get(id=1)
```
li-boxuan: What about putting it in the `setUpTestData` method?

openhub/tests/models/test_affiliated_committer_model.py
```diff
@@ -0,0 +1,90 @@
...
+
+        name = affiliated_committer._meta.get_field('name').verbose_name
```
li-boxuan: What's the point of testing those verbose names? Maybe only test fields whose verbose names are not by default?

.coafile
```diff
@@ -52,7 +52,7 @@ shell = bash
 # Do not allow the word "coala" to ensure the repository can be generalized out
 # for use by other organizations.
 files = **
-ignore += .coafile, *requirements.txt, .travis.yml, LICENSE, .moban.yaml, .moban.dt/community-gitignore.jj2, public/**, _site/**, .ci/check_moban.sh
+ignore += .coafile, *requirements.txt, .travis.yml, LICENSE, .moban.yaml, .moban.dt/community-gitignore.jj2, public/**, _site/**, .ci/check_moban.sh, **/tests/**
```
jayvdb: No.

Stop editing this line.

You have functions which give you the name of the org, which you can use.

openhub/tests/models/test_affiliated_committer_model.py
sks444: Not working, we can't use the name `affiliated_committer` in this method.

openhub/tests/models/test_affiliated_committer_model.py
sks444: Yeah, I just wanted to make sure whenever the name of any field change, there should be a change in tests also :P

jayvdb:
Ah this has the same problem as https://github.com/coala/community/pull/152 , so we may as well block this one until that problem has been solved in 152

Issue #154, opened 2018-06-21
webservices is down and breaking everything


jayvdb:
The community website shouldnt fail if the webservices component is dead.

We should take advantage of it being down to fix our code so that it doesnt bring down the entire system.

sks444:
This is happening because of [tests for `import_contributors_data`](https://github.com/coala/community/blob/5d97537bf2aa2e7914a804c15080a4ed2fa500c1/data/tests/test_management_commands.py#L15) management command;

Maybe we can think of other ways of testing management commands.

sks444:
Or maybe figure out a way to save the old data somewhere and fetch the data from there whenever the webservices goes down next time.

jayvdb:
first check if the real data from webservices has 0 records.
If it does, something went wrong and all tests of the real data should be skipped

PR #155, opened 2018-06-21
test_management_commands.py: Fix webservices error

sks444:
This checks if the real data from the webservices
has 0 records, if it does, it skips all the tests
of the real data.

Fixes https://github.com/coala/community/issues/154

### DIFF
--- a/data/tests/test_management_commands.py
+++ b/data/tests/test_management_commands.py
@@ -1,3 +1,5 @@
+import unittest
+
 from django.core.management import call_command
 from django.test import TestCase
 
@@ -12,5 +14,8 @@ class ImportContributorDataTest(TestCase):
 
     def test_command_import_contributors_data(self):
         contributors = Contributor.objects.all()
+        if not contributors:
+            raise unittest.SkipTest(
+                'No record of contributors from webservices')
         self.assertIn('jayvdb',
                       [contributor.login for contributor in contributors])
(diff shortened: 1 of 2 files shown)


data/tests/test_management_commands.py
```diff
@@ -12,5 +12,7 @@ def setUpTestData(cls):
 
     def test_command_import_contributors_data(self):
         contributors = Contributor.objects.all()
+        if not contributors:
+            return
```
jayvdb: https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures

sks444: I am having a bit of problem here:

How can we use the Contributor model outside of test methods?

Suppose I do this:
```
num_contrib = Contributor.objects.all().count()
@skipIf(num_contrib==0, 'There are not cotrib..')
def test_command_import_contributors_data(self):
     ...
```
In this case, the test will always be skipped no matter if we are getting the data from the webservices or not as we are not using the model from the `SetUpTestData`.

jayvdb: there is always the option of using `unittest.SkipTest` it is mentioned in that section.

jayvdb:
ack https://github.com/coala/community/pull/155/commits/25e0994750129581e39cff2d6e056da097535753 https://github.com/coala/community/pull/155/commits/baae22188f1880fd6fd4102a4a3f12ecce87381a

jayvdb:
@gitmate-bot ff

PR #160, opened 2018-06-28
Gamification1

sks444:


data/contrib_data.py
```diff
@@ -2,13 +2,13 @@
...
-from community.git import get_org_name
-from data.models import Contributor
+# from community.git import get_org_name
+from data.models import Contributor, Team
 
 
 def get_contrib_data():
     logger = logging.getLogger(__name__)
-    IMPORT_URL = 'https://webservices.' + get_org_name() + '.io/contrib/'
+    IMPORT_URL = 'https://pastebin.com/raw/6P6W3ziV'
```
sks444: Using fake data, until we get the data from webservices.

gamification/create.py
```diff
@@ -0,0 +1,301 @@
...
+    level_objects_list = [
+        Level(number=1, min_score=0, max_score=5, name='newbie'),
+        Level(number=2, min_score=5, max_score=10, name='beginner'),
+        Level(number=3, min_score=10, max_score=15, name='learner'),
+        Level(number=4, min_score=15, max_score=20, name='intermediate'),
+        Level(number=5, min_score=20, max_score=25, name='cool'),
+        Level(number=6, min_score=25, max_score=30, name='awesome'),
+        Level(number=7, min_score=30, max_score=35, name='master'),
+        Level(number=8, min_score=35, max_score=40, name='legend'),
+        Level(number=9, min_score=40, max_score=45, name='expert'),
```
sks444: Levels are not well calculated and finalized yet, this is assumptions to test the basic gamification system.

gamification/management/commands/import_issues_data.py
```diff
@@ -0,0 +1,18 @@
...
+
+    ISSUES = staticmethod(get_issues)
+    IMPORT_DATA = staticmethod(import_data)
+
+    def handle(self, *args, **options):
+        # for issue in self.ISSUES('github'):
+        #    self.IMPORT_DATA('github', issue)
+        # for issue in self.ISSUES('gitlab'):
+        #    self.IMPORT_DATA('gitlab', issue)
+        for issue in self.ISSUES('test'):
```
sks444: This will be removed when we will have issues data coming from the webservices.

gamification/management/commands/import_mrs_data.py
```diff
@@ -0,0 +1,18 @@
...
+
+    MRS = staticmethod(get_mrs)
+    IMPORT_DATA = staticmethod(import_data)
+
+    def handle(self, *args, **options):
+        # for mr in self.MRS('github'):
+        #    self.IMPORT_DATA('github', mr)
+        # for mr in self.MRS('gitlab'):
+        #    self.IMPORT_DATA('gitlab', mr)
+        for mr in self.MRS('test'):
```
sks444: This will be removed when we will have mrs data coming from the webservices.

sks444:
I think we I should make classes for the files `create.py` and `update.py` with the names `CreateGamificationData` and `UpdateGamificationData`  respectively with their own handler method, So that I don't have to call each of the methods of those files in the management commands explicitly.

gamification/models.py
```diff
@@ -0,0 +1,202 @@
...
+        """
+        Find the badge based on the activities peformed by the newcomer.
+
+        :param activities: a QuerySet dict containing the 'name'
+                           as key and 'name of the activity' as value
+        :return: a badge object
+        """
+        activities = [activity['name'] for activity in activities]
+        badge = Badge.objects.filter(
+            b_activities__name__in=activities).distinct()
```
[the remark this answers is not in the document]
sks444: I wonder if we could do it in one query, something like:
```
 badge = Badge.objects.filter(
            b_activities__all__name__in=activities).distinct()
```
Otherwise I am getting a long way:
```
    def find_badges_for_activity(self, activities):
        activities = [activity['name'] for activity in activities]
        badge_objects_list = []
        badges = Badge.objects.all()
        for badge in badges:
            b_activities = badge.b_activities.values('name')
            b_activities = [b_activity['name'] for b_activity in b_activities]
            match_activity_list = []
            for b_activity in b_activities:
                if b_activity in activities:
                    match_activity_list.append(1)
            if b_activities.count() == len(match_activity_list):
                badge_objects_list.append(badge)
       return badge_objects_list
```

gamification/models.py
```diff
@@ -0,0 +1,202 @@
...
+
+        new_level = self.find_level_for_score(new_score)
+        if new_level.number > self.level.number:
+            self.level = new_level
+
+    def add_activity(self, points, activity_string):
+        """
+        Add activity to the newcomer.
+
+        This methof checks if the current activity is
```
sks444: methof --> method

data/contrib_data.py
```diff
@@ -27,14 +27,23 @@ def get_contrib_data():
...
+        contributor.pop('teams')
         c, create = Contributor.objects.get_or_create(
             **contributor
         )
         if create:
             c.save()
+            for team in teams:
+                t, created = Team.objects.get_or_create(name=team)
+                if created:
+                    t.save()
```
li-boxuan: This is redundant...?

data/issues.py
```diff
@@ -0,0 +1,88 @@
...
+
+def fetch_issues(hoster):
+    """
+    Get issues opened by newcomers
+
+    :param hoster: a string representing hoster, e.g. 'github'
+    :return: a json of issues data
+    """
+    logger = logging.getLogger(__name__)
+    if hoster == 'github':
```
li-boxuan: maybe convert to lower case and then compare?

data/issues.py
```diff
@@ -0,0 +1,88 @@
...
+    author = issue.pop('author')
+    try:
+        c = Contributor.objects.get(login=author)
+        issue['author'] = c
+        issue['hoster'] = hoster
+        i, created = Issue.objects.get_or_create(
+            **issue
+            )
+        if created:
+            i.save()
```
li-boxuan: Same as previous comment, this is not needed. See https://docs.djangoproject.com/en/2.0/ref/models/querysets/#get-or-create
`If an object is not found, get_or_create() will instantiate and save a new object, returning a tuple of the new object and True`

li-boxuan: There are many other places in this pull request which explicitly call `save` method after `get_or_create`, which is redundant.

templates/gamification.html
```diff
@@ -0,0 +1,44 @@
...
+      {% for participant in participants %}
+      <div class="container">
+        <div class="row">
+          <div class="col-sm-6 col-md-4">
+            <div class="thumbnail">
+              <div class="caption">
+                <p>Username: {{ participant.username }}</p>
+                <p>Score: {{ participant.score }}</p>
+                <p>Level: {{ participant.level.name }}</p>
+                <p>Activities Perfomred:
```
Vamshi99: `Perfomred` --> `Performed`?

data/issues.py
prnvdixit: Agreed ^^

data/issues.py
```diff
@@ -0,0 +1,84 @@
...
+    Get issues opened by newcomers
+
+    :param hoster: a string representing hoster, e.g. 'github'
+    :return: a json of issues data
+    """
+    logger = logging.getLogger(__name__)
+    if hoster == 'github':
+        IMPORT_URL = web_url + 'issues/github/all'
+    elif hoster == 'gitlab':
+        IMPORT_URL = web_url + 'issues/github/all'
```
prnvdixit: Shouldn't this IMPORT_URL be for `gitlab`?

gamification/update.py
```diff
@@ -0,0 +1,90 @@
...
+            # Get activity and points based on the labels on the issue
+            points, activity_string = get_issue_activity(i_labels)
+
+            # Update participant data who opened the issue
+            ncm.add_points(points, activity_string)
+            ncm.save()
+    if mr.state == 'closed':
+        ncm = Participant.objects.get(username=mr.author)
+        points = 5
+        activity = 'Closed a merge_request without merge'
```
jayvdb: Someone else may have closed it.

"Merge request was closed without merge"

But it would be great to identifier the closer and deduct even more points if they closed their own PR! ;-)

sks444:
@prnvdixit, not as of now.
>issue being closed without a MR getting merged for it 

The reason is we don't have the data of those mrs who are not opened by newcomers.

Yes, if we can identify the closer of the issue, we can deduct points for the newcomers who closes their own issues, as it clearly indicates that the mrs for those issues aren't merged. But `IGitt` doesn't provide the details of the issue or mr closer yet.

But another way of doing it is to check the labels on the issues, and in case the labels are `invalid` or `duplicate` then deduct points of the newcomer who opened the issue.

sks444:
Netlify is failing because of the time limit.

import commands is taking so much time as there is lots of issues and mrs opened, closed or merged by current newcomers team.

I have more than 1500 mrs opened, merged and closed by newcomers and more than 200 issues opened and closed by newcomers.

`update_gamification_data` is also taking some more time as it is iterating through all the issues and mrs.

need to create a workaround for the time limit failure otherwise the pr is ready for review. And it runs fine locally. Here is the output https://gist.githubusercontent.com/sks444/50f55fb15eee7a8d9a9626805826d1a6/raw/4aad87bfcf4a2f573355a9bcd3a36c880f1aa262/gistfile1.txt

jayvdb:
Remove gitter stuff . That is a separate PR and it is not nearly ready.

gamification/models.py
```diff
@@ -0,0 +1,202 @@
...
+        :return: a badge object
+        """
+        activities = [activity['name'] for activity in activities]
+        badge = Badge.objects.filter(
+            b_activities__name__in=activities).distinct()
+        return badge
+
+    def add_badge(self, activities):
+        """
+        Add badge to newcomer based on the activities performed.
```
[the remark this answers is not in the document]
jayvdb: This is still hiding in one of your commits.  Squash your commits!

gamification/activity.py
```diff
@@ -0,0 +1,30 @@
...
+
+    # Sorting the labels to get consistent activity
+    labels_list.sort()
+    if activity_type == 'issue':
+        activity = 'Created a '
+    elif activity_type == 'merge_request':
+        activity = 'Solved a '
+    for label in labels_list:
+        activity = activity + label + ' '
+    activity = activity + 'issue'
```
jayvdb: this will append 'issue' to merge_request activities ?

gamification/create.py
```diff
@@ -0,0 +1,246 @@
+from gamification.models import (
+    Level,
+    BadgeActivity,
+    Badge,
+    )
```
jayvdb: dedent, everywhere that you have done this.

gamification/management/commands/create_gamification_data.py
```diff
@@ -0,0 +1,20 @@
...
+    create_levels,
+    create_badge_activity,
+    create_badges,
+    add_activity_to_badge,
+    )
+from gamification.newcomers import create_newcomers
+
+
+class Command(BaseCommand):
+    help = 'Create gamification data'
```
jayvdb: "data" is too generic as "data" is also used for participant data.

Typically this stuff is "config data", or "reference data"

Also put all of those files into a common subdirectory, like "gamification/init/points.py" . If you can think of an alphabetically earlier name than "init" that would be good.  Maybe "begin".  And then rename other stuff to appear after it, so that patches always include any of this reference data changes first before code changes.

Create other subdirectories as needed to group related modules together.

gamification/management/commands/create_gamification_data.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def handle(self, *args, **options):
+        create_levels()
+        create_newcomers()
```
jayvdb: move this out to a "create_participants" command

gamification/management/commands/update_gamification_data.py
```diff
@@ -0,0 +1,26 @@
...
+    update_participants_data_with_mr,
+    get_issue_objects,
+    update_participants_data_with_issue,
+    get_participant_objects,
+    award_badges,
+    )
+
+
+class Command(BaseCommand):
+    help = 'Update gamification data'
```
jayvdb: "Update participants data" and change command name accordingly

gamification/models.py
```diff
@@ -0,0 +1,206 @@
...
+            # be different for this activity
+            new_activity = Activity.objects.create(
+                name=activity_string, points=points,
+                number_of_times=previously_performed + 1,
+                performed_at=performed_at,
+                updated_at=updated_at,
+                performer=self)
+            self.activities.add(new_activity)
+        else:
+            activity = Activity.objects.create(name=activity_string,
```
jayvdb: args are aligned differently here than in the other branch of the `if`.

that makes the reader not see that they are very similar sets of code.

gamification/models.py
```diff
@@ -0,0 +1,206 @@
...
+                b_activity_words_list = b_activity.split()
+                for activity in activities:
+                    if all(x in activity for x in b_activity_words_list):
+                        match_activity_list.append(1)
+                        break
+            if len(b_activities) == len(match_activity_list):
+                # The participant have performed all the activities listed
+                # in the badge
+                badge_objects_list.append(badge)
+        return badge_objects_list
```
jayvdb: Add a blank line above this so it is clearly separated from the for loop

gamification/activity.py
sks444: Yes, `merge_request` or `issue` activities both.

If the activity is merge_request, then it would become:
```
 Solved a area/documentation difficulty/medium importance/medium issue
```
And if the activity is `issue`, then it would become:
```
 Created a area/documentation difficulty/medium importance/medium issue
```

jayvdb:
Not reviewing until after test coverage rules have been added.

gamification/data/labels.py
```diff
@@ -0,0 +1,22 @@
+def used_labels():
+    labels = [
```
jayvdb: declare this as a module level CONSTANT, using tuple syntax.

There is no benefit putting it inside a function, because they are not complex objects from the model.

gamification/data/points.py
```diff
@@ -0,0 +1,52 @@
...
+            'area/lintbears': 7,
+            'area/nativebears': 5,
+            'type/bug': 5,
+            'type/feature': 5,
+            'bear proposal': 5,
+            'status/invalid': -5,
+            'status/duplicate': -5,
+        }
+        return points_dict
+    if activity_type == 'merge_request':
```
jayvdb: elif

but again, declare these as module level CONSTANTS.

There is no benefit putting them inside a function.

gamification/process/activity.py
```diff
@@ -0,0 +1,30 @@
...
+        if label not in used_labels():
+            labels_list.remove(label)
+
+    # Sorting the labels to get consistent activity
+    labels_list.sort()
+    if activity_type == 'issue':
+        activity = 'Created a '
+        activity = activity + ' '.join(labels_list) + ' issue'
+        return activity
+    elif activity_type == 'merge_request':
```
jayvdb: can you see similarity between these two branches ?

gamification/process/update.py
```diff
@@ -0,0 +1,142 @@
...
+        mr_points, mr_activity_string = get_activity_with_points(
+            'merge_request', mr_labels)
+
+        # Update participant data
+        mr_author.add_points(mr_points,
+                             mr_activity_string,
+                             mr_performed_at,
+                             mr_updated_at,
+                             )
+        mr_author.save()
```
jayvdb: you can save the author at the end of the function, outside the branch.

gamification/process/update.py
```diff
@@ -0,0 +1,142 @@
...
+        try:
+            issues = mr.get_closes_issues_object()
+        except Exception as ex:
+            logger.error(ex)
+            return
+        for issue in issues:
+            issue_labels = issue.labels.values('name')
+
+            # Get the participant who opened the issue
+            issue_author = Participant.objects.get(username=issue.author)
```
jayvdb: first check if mr_author is who opened the issue

(ideally this would first look for any authors who authored multiple issues that are being closed, and grouped those issues together before writing that updated author to the db)

gamification/process/update.py
```diff
@@ -0,0 +1,142 @@
...
+            issue_points, issue_activity_string = get_activity_with_points(
+                'issue', issue_labels)
+
+            # Update participant data who opened the issue
+            issue_author.add_points(issue_points,
+                                    issue_activity_string,
+                                    issue_performed_at,
+                                    issue_updated_at,
+                                    )
+            issue_author.save()
```
jayvdb: if issue_author is mr_author, do not save , as it will be saved at the end of the function, and we dont want to do a lot of db updates unnecessarily

gamification/process/update.py
```diff
@@ -0,0 +1,142 @@
...
+    labels in the issue and based on that it based on that it
+    creates an activity and deduct 5 points.
+    """
+    labels = issue.labels.values('name')
+    labels_list = [label['name'] for label in labels]
+    issue_author = Participant.objects.get(username=issue.author)
+
+    if 'status/invalid' in labels_list or 'status/duplicate' in labels_list:
+        issue_points = -5
+        issue_activity = 'Created a invalid or duplicate issue'
```
jayvdb: you can improve this activity description using the actual label which was encountered, and only when both appear, deciding which is more important in the label.

other activities use the convention

> 'Created a difficulty/newcomer type/bug issue'

here the same should happen.

gamification/process/update.py
```diff
@@ -0,0 +1,142 @@
...
+
+    This method checks for 'status/invalid' or 'status/duplicate'
+    labels in the issue and based on that it based on that it
+    creates an activity and deduct 5 points.
+    """
+    labels = issue.labels.values('name')
+    labels_list = [label['name'] for label in labels]
+    issue_author = Participant.objects.get(username=issue.author)
+
+    if 'status/invalid' in labels_list or 'status/duplicate' in labels_list:
```
jayvdb: these two constant labels are config data and belong in the data submodule.

gamification/data/points.py
```diff
@@ -0,0 +1,52 @@
+def get_points(activity_type):
+    """
+    Get points_dict for issue and merge_request.
+
+    This method contains the points assigned to the activity
+    variables, which will be used throughout the gamification
+    system.
+    """
+    if activity_type == 'issue':
+        points_dict = {
```
jayvdb: these dicts are basic data types, and so again can be module level CONSTANTS.

gamification/process/update.py
sks444: yes but then first I have to use:
```
else:
    return
```
As  when `mr.state` will be `open` I will get an error:
```
UnboundLocalError: local variable 'mr_points' referenced before assignment
```
So, is it okay to use `else`?

gamification/process/activity.py
sks444: yes, similar case ad https://github.com/coala/community/pull/160#discussion_r208126754

gamification/process/update.py
jayvdb: no.  move your variable to the top

gamification/process/update.py
```diff
@@ -0,0 +1,140 @@
...
+            logger.error(ex)
+            return
+        for issue in issues:
+            issue_labels = issue.labels.values('name')
+            issue_performed_at = issue.created_at
+            issue_updated_at = issue.updated_at
+            # Get activity and points based on the labels on the issue
+            issue_points, issue_activity_string = get_activity_with_points(
+                'issue', issue_labels)
+            issue_author = Participant.objects.get(username=issue.author)
```
jayvdb: did you check if the issue_author is the mr_author ??

sks444: yes, below.

templates/gamification.html
```diff
@@ -0,0 +1,44 @@
...
+  <body>
+    <h1>The gamification leaderboard</h1>
+    <ul>
+      {% for participant in participants %}
+      <div class="container">
+        <div class="row">
+          <div class="col-sm-6 col-md-4">
+            <div class="thumbnail">
+              <div class="caption">
+                <p>Username: {{ participant.username }}</p>
```
jayvdb: add links to github and gitlab and openhub profile ?

gamification/process/update.py
jayvdb: please re-read last comments.  you havent done this correctly. my guess is the following below is buggy, or django may recover sanely but it is dodgy as

sks444: @jayvdb, does following look good?
```
            if issue.author == mr.author:
                mr_author.add_points(issue_points,
                                     issue_activity_string,
                                     issue_performed_at,
                                     issue_updated_at,
                                     )
            else:
                issue_author = Participant.objects.get(username=issue.author)
                # Update participant data who opened the issue
                issue_author.add_points(issue_points,
                                        issue_activity_string,
                                        issue_performed_at,
                                        issue_updated_at,
                                        )
                issue_author.save()
```

templates/gamification.html
```diff
@@ -0,0 +1,44 @@
...
+                <p>Score: {{ participant.score }}</p>
+                <p>Level: {{ participant.level.name }}</p>
+                <p>Activities Performed:
+                {% for activity in participant.activities.all %}
+                  <p>{{ forloop.counter }}. {{ activity.name }}, number_of_times:
+                    {{ activity.number_of_times }}, performed_at:
+                    {{ activity.performed_at }} updated_at: {{ activity.updated_at }}
```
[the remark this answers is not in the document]
sks444: Actually, this is directly dumped from the database.

But, I think it has time zone support:
```
In [8]: Activity.objects.get(id=432)
Out[8]: <Activity: Solved a difficulty/newcomer type/bug issue>

In [9]: a = Activity.objects.get(id=432)

In [10]: a.performed_at
Out[10]: datetime.datetime(2016, 2, 21, 5, 4, 25, tzinfo=<UTC>)
```

gamification/process/update.py
jayvdb: that looks better

templates/gamification.html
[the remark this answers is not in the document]
jayvdb: so you need to render it so that the reader knows the timezone it is in.

if it is always in UTC, a simple note at the top would be ok.

gamification/data/labels.py
```diff
@@ -0,0 +1,51 @@
...
+    'bear/proposal',
+)
+
+NEGATIVE_POINT_LABELS = (
+    'status/invalid',
+    'status/duplicate',
+)
+
+
+def get_sorted_labels(labels):
```
jayvdb: this isnt 'data'.

it belongs in a different subcomponent.

gamification/data/participants.py
```diff
@@ -0,0 +1,13 @@
+from data.newcomers import active_newcomers
```
jayvdb: this isnt 'data'.

it belongs in a different subcomponent.

gamification/data/points.py
```diff
@@ -0,0 +1,61 @@
...
+    'area/nativebears': 7,
+    'type/bug': 10,
+    'type/feature': 7,
+    'bear proposal': 5,
+    'status/invalid': -5,
+    'status/duplicate': -5,
+}
+
+
+def get_total_points(activity_type, labels):
```
jayvdb: this isnt 'data'.

it belongs in a different subcomponent.

gamification/models.py
```diff
@@ -0,0 +1,215 @@
...
+            previously_performed = Activity.objects.filter(
+                name=activity_string,
+                performer=self.username).count()
+
+            # Creating a new activity with increasing the number_of_times
+            # With one because the performed_at and updated_at time would
+            # be different for this activity
+            new_activity = Activity.objects.create(
+                name=activity_string, points=points,
+                number_of_times=previously_performed + 1,
```
jayvdb: And I am a little unsure about why this creates a additional activity with `number_of_times = .. ` if the old activity is being kept.  That would mean that the sum of activities and their `number_of_times` would double count the number of activities performed.  Which part of the cEP needed this approach?  Pls lets discuss.

----

Assuming it is necessary/ appropriate to have a new activity when one already exists,...

This and the `else:` branch below do the same thing, except for the `number_of_times=previously_performed + 1` which can be a variable initialised to `number_of_times=0` and revised in the logic branches.

The creation of the new activity doesnt need to be inside the if/else.

gamification/process/activity.py
```diff
@@ -0,0 +1,17 @@
...
+                          activity. e.g. issue or merge_request.
+    :param labels:        a list of sorted labels.
+    :return:              a touple of points and activity string.
+    """
+    if activity_type == 'issue':
+        activity = 'Created a '
+    elif activity_type == 'merge_request':
+        activity = 'Solved a '
+    else:
+        return
```
jayvdb: raise an exception.

gamification/labels.py
```diff
@@ -0,0 +1,51 @@
...
+    'bear/proposal',
+)
+
+NEGATIVE_POINT_LABELS = (
+    'status/invalid',
+    'status/duplicate',
+)
+
+
+def get_sorted_labels(labels):
```
jayvdb: this *function* should not be in gamification/data , but the data above should be in gamification/data !

gamification/points.py
```diff
@@ -0,0 +1,62 @@
+ISSUE_POINTS_DICT = {
```
jayvdb: this is data.  it goes in `gamification/data/`

gamification/models.py
jayvdb: Decision is to keep the field `number_of_times` in the model atm, but not use it.
A note can be added in the model that it is unused.

We can tidy it up later.

gamification/models.py
```diff
@@ -0,0 +1,208 @@
...
+        :return:           a badge object.
+        """
+        activities = [activity['name'] for activity in activities]
+        badge_objects_list = []
+        badges = Badge.objects.all()
+        for badge in badges:
+            b_activities = badge.b_activities.values('name')
+            b_activities = [b_activity['name'] for b_activity in b_activities]
+            # Skip the badge which have no activities
+            if len(b_activities) == 0:
```
jayvdb: `if not len(b_activities):`

gamification/points.py
```diff
@@ -0,0 +1,62 @@
...
+    :return:              a integer value representing the total points.
+    """
+    points = 0
+    for label in labels:
+        if activity_type == 'issue':
+            plus_points = ISSUE_POINTS_DICT.get(label, 0)
+        elif activity_type == 'merge_request':
+            plus_points = MR_POINTS_DICT.get(label, 0)
+        else:
+            raise UnboundLocalError('Activity is not defined.')
```
jayvdb: do not misuse exceptions.

that one has a specific meaning, and indicates there was a Python syntax error ; it is extremely unlikely you should *ever* raise that one.

here, you need `ValueError`, as the value of the arg was unusable / unexpected.

gamification/points.py
```diff
@@ -0,0 +1,62 @@
...
+def get_total_points(activity_type, labels):
+    """
+    Get total points.
+
+    :param activity_type: a string representing the type of
+                          the activity. e.g. issue or merge_request
+    :param labels:        a list of sorted labels.
+    :return:              a integer value representing the total points.
+    """
+    points = 0
```
jayvdb: also define `plus_points` here, so that a `UnboundLocalError` can not occur.

An initial value of 0 is sensible here

gamification/process/activity.py
```diff
@@ -0,0 +1,17 @@
...
+                          activity. e.g. issue or merge_request.
+    :param labels:        a list of sorted labels.
+    :return:              a touple of points and activity string.
+    """
+    if activity_type == 'issue':
+        activity = 'Created a '
+    elif activity_type == 'merge_request':
+        activity = 'Solved a '
+    else:
+        raise UnboundLocalError('Activity is not defined.')
```
jayvdb: `UnboundLocalError` is very wrong.  You need to define the variable `activity` before logic branches to avoid this exception.  Define it with an initial value of `None`

gamification/tests/activityTest.py
```diff
@@ -0,0 +1,32 @@
...
+        activity = get_activity('merge_request', labels)
+
+        # There are two labels 'difficulty/newcomer' and 'type/bug'
+        # and the activity_type is mr, so the activity should be
+        expected_activity = 'Solved a difficulty/newcomer type/bug issue'
+        self.assertEquals(activity, expected_activity)
+
+    def test_undefined_activity(self):
+        labels = ['difficulty/newcomer', 'type/bug']
+        with self.assertRaises(UnboundLocalError):
```
jayvdb: you'll need to change this to ValueError

gamification/models.py
```diff
@@ -0,0 +1,208 @@
...
+    performer = models.ForeignKey('Participant',
+                                  on_delete=models.CASCADE,
+                                  null=True)
+    # The date and time this activity was performed
+    performed_at = models.DateTimeField(null=True)
+
+    # The date and time this activity was last updated
+    updated_at = models.DateTimeField(null=True)
+
+    # Number of questions asked on gitter rooms while
```
jayvdb: remove this; add gitter stuff in the gitter PR

gamification/models.py
```diff
@@ -0,0 +1,208 @@
...
+    def update_score_and_level(self, points):
+        """
+        Update score and level based on points.
+        """
+        if points < 0 and self.score < abs(points):
+            self.score = 0
+            new_score = self.score
+        else:
+            self.score += points
+            new_score = self.score
```
jayvdb: `new_score` is an unnecessary variable

gamification/models.py
```diff
@@ -0,0 +1,208 @@
...
+
+        new_level = self.find_level_for_score(new_score)
```
jayvdb: `new_level` is an unnecessary variable

gamification/data/labels.py
```diff
@@ -0,0 +1,23 @@
...
+    'difficulty/high',
+    'importance/low',
+    'importance/medium',
+    'importance/high',
+    'importance/critical',
+    'area/documentation',
+    'area/upstream',
+    'area/genericbears',
+    'area/lintbears',
+    'area/nativebears',
```
jayvdb: the lists in this file are not necessary.  They can be derived from the data in `points.py` .

DRY

To reduce change to your algorithms, create them as CONSTANTS in `gamification/labels.py` populated via a algorithm rather than static data

gamification/management/commands/update_participants_data.py
```diff
@@ -0,0 +1,26 @@
...
+    award_badges,
+)
+
+
+class Command(BaseCommand):
+    help = 'Update participants data'
+
+    def handle(self, *args, **options):
+        for mr in get_mr_objects():
+            if mr.labels.count() == 0:
```
jayvdb: `if not mr.labels.count():`  (or maybe `if not mr.labels:` ? im not sure if django supports that) 

and this must be logged as a warning in our logging system.

someones work wasnt counted, and was silently discarded!!!

likewise with issues below.

gamification/process/activity.py
```diff
@@ -0,0 +1,18 @@
...
+    Get an activity based on the labels on issues or mrs.
+
+    :param activity_type: a string representing the type of the
+                          activity. e.g. issue or merge_request.
+    :param labels:        a list of sorted labels.
+    :return:              a touple of points and activity string.
+    """
+    activity = None
+    if activity_type == 'issue':
+        activity = 'Created a '
```
jayvdb: qactivity is a bad name for this; it is a prefix

gamification/process/update.py
```diff
@@ -0,0 +1,143 @@
...
+            return
+        for issue in issues:
+            issue_labels = issue.labels.values('name')
+            issue_performed_at = issue.created_at
+            issue_updated_at = issue.updated_at
+            # Get activity and points based on the labels on the issue
+            issue_points, issue_activity_string = get_activity_with_points(
+                'issue', issue_labels)
+            if issue.author == mr.author:
+                mr_author.add_points(issue_points,
```
jayvdb: `issue_author = mr.author`

then move the `add_points()` to after the `if`

gamification/process/update.py
```diff
@@ -0,0 +1,143 @@
...
+                issue_author = Participant.objects.get(username=issue.author)
+                # Update participant data who opened the issue
+                issue_author.add_points(issue_points,
+                                        issue_activity_string,
+                                        issue_performed_at,
+                                        issue_updated_at,
+                                        )
+
+    elif mr.state == 'closed':
+        mr_points = -5
```
jayvdb: this should be defined in gamification/data/points.py

gamification/tests/activityTest.py
```diff
@@ -0,0 +1,32 @@
+from django.test import TestCase
+from gamification.process.activity import get_activity
+
+
+class GetActivityTest(TestCase):
+
+    @classmethod
+    def setUpTestData(cls):
```
jayvdb: not used; then dont include it

gamification/models.py
```diff
@@ -0,0 +1,182 @@
...
+        :return:           a badge object.
+        """
+        activities = [activity['name'] for activity in activities]
+        badge_objects_list = []
+        badges = Badge.objects.all()
+        for badge in badges:
+            b_activities = badge.b_activities.values('name')
+            b_activities = [b_activity['name'] for b_activity in b_activities]
+            # Skip the badge which have no activities
+            if not len(b_activities):
```
jayvdb: which badges have no activities?

I guess this should at least be a 'info' level log entry

gamification/models.py
```diff
@@ -0,0 +1,182 @@
...
+                continue
+            match_activity_list = []
+            for b_activity in b_activities:
+                b_activity_words_list = b_activity.split()
+                for activity in activities:
+                    if all(x in activity for x in b_activity_words_list):
+                        match_activity_list.append(1)
+                        break
+            if len(b_activities) == len(match_activity_list):
+                # The participant have performed all the activities listed
```
jayvdb: add a log entry that the participant was awarded a badge

gamification/management/commands/update_participants_data.py
```diff
@@ -0,0 +1,33 @@
...
+        for mr in get_mr_objects():
+            if not mr.labels.count():
+                logger.warning(
+                    'Merge request: %s has no labels.' % mr)
+                continue
+            update_participants_data_with_mr(mr)
+        for issue in get_issue_objects():
+            if not issue.labels.count():
+                logger.warning(
+                    'Issue: %s has no labels.' % issue)
```
jayvdb: please look at the log https://deploy-preview-160--coala-community.netlify.com/log/

that is really hard to investigate these warnings.  Can we have a URL in the log ?

gamification/process/update.py
```diff
@@ -0,0 +1,140 @@
...
+            return
+        for issue in issues:
+            issue_labels = issue.labels.values('name')
+            issue_performed_at = issue.created_at
+            issue_updated_at = issue.updated_at
+            # Get activity and points based on the labels on the issue
+            issue_points, issue_activity_string = get_activity_with_points(
+                'issue', issue_labels)
+            if issue.author == mr.author:
+                issue_author = mr_author
```
jayvdb: `mr_author` is invalid.

and the fact that this didnt cause a build break worries me.

does that mean we have zero mrs (by newcomers) where the MR author is the same as the issue author?

I might be wrong, but my guess is that at least @harshhx17 has created an issue and also worked on it.

gamification/models.py
jayvdb: oh wow, this log entry is quite noisy as it is being done many times.

drop it down to debug, and add a comment in data/badges.py about the two badges without activities.

gamification/process/update.py
jayvdb: ahhh .. `issue.author == mr.author` and `issue_author = mr_author` is confusing because of the `.` vs `_`.

Use `if issue.author == mr_author.username:` ?

gamification/management/commands/update_participants_data.py
sks444: Yes, but then we have to create a function to generate the url from the issue and mr objects *as we don't have an url field in the Issue or MergeRequest models*.
e.g. 
```
url = 'https://' + issue.hoster.lower() +  '/' + issue.repo + '/issues/' + issue.number
```
this will work for both the hoster but for a merge_request url depend on the hoster as GitHub-->pull is change to `merge_request` on GitLab.

Though, we have a `repo` field in the models but currently `issue.repo` would return `null` But I will change the api to include `repo` key soon.

gamification/process/update.py
sks444: Strange that we are not getting coverage when we use `if issue.author == mr_author.username:` but `issue.author == mr.author` the full coverage.

jayvdb: ok, not worth the pain of debugging this atm

Issue #167, opened 2018-07-19
import_contributors_data is failing


sks444:
```  
File "/home/travis/build/coala/community/data/tests/test_management_commands.py", line 13, in setUpTestData
    call_command('import_contributors_data')
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/django/core/management/__init__.py", line 131, in call_command
    return command.execute(*args, **defaults)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/django/core/management/base.py", line 330, in execute
    output = self.handle(*args, **options)
  File "/home/travis/build/coala/community/data/management/commands/import_contributors_data.py", line 13, in handle
    for contributor in self.CONTRIBUTORS():
TypeError: 'NoneType' object is not iterable
```
The error is happenning while executing [get_contrib_data](https://github.com/coala/community/blob/master/data/contrib_data.py#L9) method; `requests.get` method in the [line 14](https://github.com/coala/community/blob/master/data/contrib_data.py#L14) is not returning any response, that's why the except statement is giving following error and returning `None` and we are getting `TypeError: 'NoneType' object is not iterable`.
```
SSLError: HTTPSConnectionPool(host='webservices.coala.io', port=443): Max retries exceeded with url: /contrib/ (Caused by SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:645)'),))
```

PR #169, opened 2018-07-23
data/: Fix import_contributors_data command

sks444:
This fix the import_contributors_data command
when the webservices has a server error and our
returned data by the method get_contrib_data is None,
then while iterating through the contributors data
in the management command we need to first check
if the data is not empty to avoid the error:
```
TypeError: 'NoneType' object is not iterable
```
Fixes https://github.com/coala/community/issues/167

### DIFF
--- a/data/management/commands/import_contributors_data.py
+++ b/data/management/commands/import_contributors_data.py
@@ -10,5 +10,7 @@ class Command(BaseCommand):
     IMPORT_DATA = staticmethod(import_data)
 
     def handle(self, *args, **options):
-        for contributor in self.CONTRIBUTORS():
-            self.IMPORT_DATA(contributor)
+        CONTRIBUTORS_DATA = self.CONTRIBUTORS()
+        if CONTRIBUTORS_DATA:
+            for contributor in CONTRIBUTORS_DATA:
+                self.IMPORT_DATA(contributor)


data/management/commands/import_contributors_data.py
```diff
@@ -10,5 +10,6 @@ class Command(BaseCommand):
     IMPORT_DATA = staticmethod(import_data)
 
     def handle(self, *args, **options):
-        for contributor in self.CONTRIBUTORS():
-            self.IMPORT_DATA(contributor)
+        if self.CONTRIBUTORS():
```
jayvdb: @sks444 , wouldnt this apply to all of the import commands, if the data is missing / empty?

dont call `self.CONTRIBUTORS()` twice

sks444: In the OpenHub commands, we won't get the error `TypeError: 'NoneType' object is not iterable` if the data is missing/empty.

Because in the data app we are [returning `None` explicitly](https://github.com/coala/community/blob/master/data/contrib_data.py#L21) when there is any problem in getting the data from the url.

But in the OpenHub commands, we will get the 'empty list' for the same.

data/management/commands/import_contributors_data.py
```diff
@@ -10,5 +10,6 @@ class Command(BaseCommand):
     IMPORT_DATA = staticmethod(import_data)
 
     def handle(self, *args, **options):
-        for contributor in self.CONTRIBUTORS():
-            self.IMPORT_DATA(contributor)
+        if get_contrib_data():
```
jayvdb: `get_contrib_data()` is called and does some work ... then you delete it by not storing it in a variable!!

data/management/commands/import_contributors_data.py
jayvdb: The point was you are calling `self.CONTRIBUTORS()` twice.

It does work.

Dont waste precious computation time by doing it twice.

jayvdb:
@gitmate-bot ff

Issue #165, opened 2018-07-19
gitorg.py: Unused variable


MacBox7:
Pyflakes error report =>
```
./gci/gitorg.py:55: local variable 'e' is assigned to but never used
```
difficulty/newcomer

Pawank221:
please assign me this issue.

MacBox7:
Comment of `coala` group on gitter for assignment.

Pranav1999:
Assign me this issue

PR #170, opened 2018-07-27
Gitorg.py: Remove unused variable

cypher-naman:
This removes an unused variable 'e' from the code.

fixes https://github.com/coala/community/issues/165

### DIFF
--- a/gci/gitorg.py
+++ b/gci/gitorg.py
@@ -53,7 +53,7 @@ def get_issue(url):
     try:
         issue = repo.get_issue(int(issue_number))
     except Exception as e:
-        logger.error('Unable to load issue %s' % url)
+        logger.error('Unable to load issue %s: %s' % (url, e))
         return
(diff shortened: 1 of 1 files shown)


gci/gitorg.py
```diff
@@ -52,7 +52,7 @@ def get_issue(url):
         return
     try:
         issue = repo.get_issue(int(issue_number))
-    except Exception as e:
+    except Exception:
```
jayvdb: log the exception in the next line

MacBox7:
@cypher-naman please do the recommended changes.

gci/gitorg.py
cypher-naman: How and where should I log them?

MacBox7: `e` holds the exception. Don't remove it, log it. `logger.error` in the next line does that.

Issue #172, opened 2018-07-28
.travis.yml: Disabe AUTH=true job for PRs 


jayvdb:
Travis PR builds do not get tokens, so we have effectively two identical jobs running for PRs.

We do want two jobs running for builds on forks, and for master builds.

See coala core .travis.yml for how to disable jobs on PRs

PR #180, opened 2018-08-09
.travis.yml: Disable AUTH=true job for PRs

jayvdb:
The AUTH=true job was in fact the deploy job.
Pull Requests do not get auth tokens, so they can not
run that job.

Fixes https://github.com/coala/community/issues/172


jayvdb:
yay. on travis this PR only has two jobs.  compare with https://travis-ci.org/jayvdb/community/builds/413967857 which has all three jobs

Issue #181, opened 2018-08-11
Add meta-review to Gamification app


jayvdb:
We have both meta-review and gamification.

And the rules for 'Developer' just changed to require use of meta-review, so that needs to be added to gamification

https://github.com/coala/coala/commit/42a3ffac059f279cd65c9539b8e44bf222782773

shikharvaish28:
![](https://github.trello.services/images/mini-trello-icon.png) [coala/community](https://trello.com/c/r0ehK2b0/2-coala-community)

sks444:
@li-boxuan, which field in the [`participant model`](https://github.com/coala/community/blob/master/meta_review/models.py#L4) would tell us if the participant has done a meta-review or not?

li-boxuan:
If user has done a meta-review: https://github.com/coala/community/blob/master/meta_review/models.py#L23
and https://github.com/coala/community/blob/master/meta_review/models.py#L32

If user has received a meta-review:
https://github.com/coala/community/blob/master/meta_review/models.py#L17
and https://github.com/coala/community/blob/master/meta_review/models.py#L26

shikharvaish28:
What all needs to be added from the meta-review app?
Example: 
From [Participant Model](https://github.com/coala/community/blob/master/meta_review/models.py#L4), do I have to import and display fields like ```neg_in``` and ```pos_in``` ( meta-view received) and similarly for meta-review done. @sks444 @jayvdb

shikharvaish28:
Wrote this piece in ```gamification/data/config.py``` 

```
    # Adding activities to the all-rounder badge
    all_rounder_badge = Badge.objects.get(
        name='The All-Rounder')
    all_rounder_activity1 = BadgeActivity.objects.get(
        name='Solved a difficulty/newcomer issue')
    all_rounder_activity2 = BadgeActivity.objects.get(
        name='Solved a difficulty/low issue')
    all_rounder_activity3 = Participant.objects.all()
    all_rounder_activities_list = [
        all_rounder_activity1,
        all_rounder_activity2,
        all_rounder_activity3
    ]
```

Expected this to work but receiving the following traceback:

```
rm: db.sqlite3: No such file or directory
+ mkdir private _site public
+ META_REVIEW_DATA=meta_review.json
+ EXPORTED_DATA='static/tasks.yaml static/instances.yaml static/meta_review.json'
+ ISSUES_JSON=issues.json
[...]

PR #190, opened 2018-09-09
Add meta-review to gamification

shikharvaish28:
From this PR, we add the meta-review activity on the gamification page of the active newcomers.

Closes https://github.com/coala/community/issues/181

### DIFF
--- a/gamification/data/config.py
+++ b/gamification/data/config.py
@@ -76,6 +76,8 @@ def create_badge_activities():
             name='Solved a area/genericbears bear proposal issue'),
         BadgeActivity(
             name='Solved a area/lintbears bear proposal issue'),
+        BadgeActivity(
+            name='Did a meta-review or received a meta-review'),
     ]
     BadgeActivity.objects.bulk_create(b_activity__object_list)
 
@@ -241,9 +243,12 @@ def add_activities_to_badges():
         name='Solved a difficulty/newcomer issue')
     all_rounder_activity2 = BadgeActivity.objects.get(
         name='Solved a difficulty/low issue')
+    all_rounder_activity3 = BadgeActivity.objects.get(
+        name='Did a meta-review or received a meta-review')
     all_rounder_activities_list = [
         all_rounder_activity1,
         all_rounder_activity2,
+        all_rounder_activity3,
     ]
     all_rounder_badge.b_activities.add(
         *all_rounder_activities_list)
--- a/gamification/management/commands/update_participants_data.py
+++ b/gamification/management/commands/update_participants_data.py
@@ -9,6 +9,8 @@ from gamification.process.update import (
     update_participants_data_with_issue,
     get_participant_objects,
     award_badges,
+    get_meta_review_participant_objects,
+    update_participants_data_with_meta_review,
 )
 
 
@@ -31,3 +33,5 @@ class Command(BaseCommand):
             update_participants_data_with_issue(issue)
         for participant in get_participant_objects():
             award_badges(participant)
+        for participant in get_meta_review_participant_objects():
+            update_participants_data_with_meta_review(participant.login)
--- a/gamification/process/update.py
+++ b/gamification/process/update.py
@@ -4,12 +4,14 @@ from data.models import (
     MergeRequest,
     Issue,
 )
+from data.newcomers import active_newcomers
 from gamification.process.activity_points import (
     get_activity_with_points
 )
 from gamification.labels import NEGATIVE_POINT_LABELS
 from gamification.data.points import MERGE_REQUEST_CLOSED_WITHOUT_MERGE
 from gamification.models import Participant
+from meta_review.models import Reaction, Participant as meta_review_participants
 
 
 def get_mr_objects():
@@ -133,6 +135,45 @@ def update_participants_data_with_issue(issue):
 def get_participant_objects():
     participants = Participant.objects.all()
     return participants
+
+
+def get_meta_review_participant_objects():
+    return meta_review_participants.objects.all()
+
+
+meta_review_completed_list = []
+
+
+def update_participants_data_with_meta_review(meta_review_participant):
+    """
+    Update participants based on meta-review
+
+    This method updates every participant based on the meta-review
+    received or given. If the  meta-review participant is
+    in the active newcomers list then it will check if the meta-review
+    is complete or not and update the activity accordingly.
+    """
+    active_newcomers_list = active_newcomers()
+    if (meta_review_participant.login in active_newcomers_list and
+            meta_review_participant.login not in meta_review_completed_list):
+        # Get the corresponding gamification participant
+        gamification_participant = Participant.objects.get(
+            username=meta_review_participant.login)
+        reaction = Reaction.objects.get(id=meta_review_participant.login)
+        if (meta_review_participant.pos_in > 0 or
+                meta_review_participant.pos_out > 0 or
+                meta_review_participant.neg_in > 0 or
+                meta_review_participant.neg_out > 0):
+            # Add a meta-review activity to the participant
+            activity = 'Did a meta-review or received a meta-review'
+            created_at = reaction.created_at
+            updated_at = None
+            gamification_participant.add_activity(0, activity, created_at,
+                                                  updated_at)
+            meta_review_completed_list.append(meta_review_participant.login)
+            logger = logging.getLogger(__name__)
+            logger.info(meta_review_participant.login,
+                        ' has completed the meta-review activity')
 
 
 def award_badges(participant):
--- a/gamification/tests/test_models.py
+++ b/gamification/tests/test_models.py
@@ -1,5 +1,7 @@
 from django.test import TestCase
 
+from data.newcomers import active_newcomers
+
 from gamification.models import (
     Activity,
     Level,
@@ -7,6 +9,12 @@ from gamification.models import (
     Badge,
     Participant,
 )
+
+from gamification.process.update import (
+    update_participants_data_with_meta_review
+)
+
+from meta_review.models import Reaction, Participant as meta_review_participant
 
 
 class ActivityModelTest(TestCase):
@@ -341,3 +349,64 @@ class ParticipantModelTest(TestCase):
 
         # After applying add_badge method
         self.assertEquals(test_participant.badges.count(), 0)
+
+    def test_update_participants_data_with_meta_review_complete(self):
+        # Test the participant who has completed the meta-review activity
+        active_newcomers_list = active_newcomers()
+        active_newcomers_list.extend(['test2', 'test3'])
+        # meta-review participant
+        meta_review_participant.objects.create(login='test2')
+        test_meta_review_participant = meta_review_participant.objects.get(
+                                       login='test2')
+
+        Reaction.objects.create(id='test2')
+        reaction = Reaction.objects.get(
+            id=test_meta_review_participant.login)
+        # corresponding gamification participant for 'test2'
+        Participant.objects.create(username='test2')
+        test_gamification_participant = Participant.objects.get(
+                                        username='test2')
+
+        test_meta_review_participant.pos_in = 2
+        reaction.created_at = '2018-10-25 14:30:59'
+        # Before applying update_participants_data_with_meta_review method
+        self.assertEquals(test_gamification_participant.activities.count(), 0)
+        # After applying update_participants_data_with_meta_review method
+        update_participants_data_with_meta_review(test_meta_review_participant)
+        self.assertEquals(test_gamification_participant.activities.count(), 1)
+        self.assertEquals(
+            test_gamification_participant.activities.all()[0].name,
+            'Did a meta-review or received a meta-review')
+        # Check if multiple activities are not adding up for the same
+        # participant 'test2' already has meta-review added so check
+        # if activity is not added again
+        update_participants_data_with_meta_review(test_meta_review_participant)
+        self.assertEquals(test_gamification_participant.activities.count(), 1)
+
+    def test_update_participants_data_with_meta_review_incomplete(self):
+        # Test the participant who has zero meta-review activity
+        meta_review_participant.objects.create(login='test3')
+        test_meta_review_participant = meta_review_participant.objects.get(
+                                        login='test3')
+        Reaction.objects.create(id='test3')
+        reaction = Reaction.objects.get(
+            id=test_meta_review_participant.login)
+        # The corresponding gamification participant
+        Participant.objects.create(username='test3')
+        test_gamification_participant = Participant.objects.get(
+                                         username='test3')
+
+        reaction.created_at = '2018-10-25 14:30:59'
+        update_participants_data_with_meta_review(test_meta_review_participant)
+        self.assertEquals(test_gamification_participant.activities.count(), 0)
+
+        # if (meta_review_participant.login in active_newcomers_list):  is false
+        meta_review_participant.objects.create(login='test4')
+        test_meta_review_participant = meta_review_participant.objects.get(
+                                        login='test4')
+        # The corresponding gamification participant for 'test4'
+        Participant.objects.create(username='test4')
+        test_gamification_participant = Participant.objects.get(
+                                         username='test4')
+        update_participants_data_with_meta_review(test_meta_review_participant)
+        self.assertEquals(test_gamification_participant.activities.count(), 0)
--- a/gamification/tests/test_management_commands.py
+++ b/gamification/tests/test_management_commands.py
@@ -24,10 +24,10 @@ class CreateConfigDataTest(TestCase):
         self.assertEquals(levels.count(), 8)
 
         b_activities = BadgeActivity.objects.all()
-        # There must be 26 badge activities created
-        # as there are 26 badge_activities listed in
+        # There must be 27 badge activities created
+        # as there are 27 badge_activities listed in
         # file config.py
-        self.assertEquals(b_activities.count(), 26)
+        self.assertEquals(b_activities.count(), 27)
 
         badges = Badge.objects.all()
         # There must be 8 badges created
@@ -61,7 +61,7 @@ class CreateConfigDataTest(TestCase):
 
         # The All-Rounder badge must have two activities
         badge7 = Badge.objects.get(name='The All-Rounder')
-        self.assertEquals(badge7.b_activities.count(), 2)
+        self.assertEquals(badge7.b_activities.count(), 3)
 
 
 class CreateParticipantsTest(TestCase):
@@ -97,4 +97,4 @@ class UpdateParticipantsTest(TestCase):
         self.assertEquals(current_level, 'Intermediate-II')
 
         number_of_badges = participant.badges.all().count()
-        self.assertEquals(number_of_badges, 2)
+        self.assertEquals(number_of_badges, 1)


shikharvaish28:
Netlify deploy is working on my end of another branch but not on this one. :(

gamification/process/update.py
```diff
@@ -56,7 +57,7 @@ def update_participants_data_with_mr(mr):
 
         # Update participant data
         mr_author.add_points(mr_points,
-                             mr_activity_string,
+                             mr_activity_string,  # Ignore PyFlakesBear
```
li-boxuan: Any reason for this change?

gamification/process/update.py
```diff
@@ -135,6 +136,29 @@ def get_participant_objects():
     return participants
 
 
+def update_participants_data_with_meta_review(participants):
```
li-boxuan: On line 35 (https://github.com/coala/community/pull/190/files#diff-0bee1776ac8d509157f197b459b8b9afR35) you are calling `update_participants_data_with_meta_review(participant)` but here it is called `participants`. So should the argument be one participant or many participants?

gamification/process/update.py
```diff
@@ -135,6 +136,29 @@ def get_participant_objects():
...
+    """
+    pos_in_meta_review = meta_review_participants.objects.get(
+                        pos_in=participants.pos_in)
+    pos_out_meta_review = meta_review_participants.objects.get(
+                        pos_out=participants.pos_out)
+    neg_in_meta_review = meta_review_participants.objects.get(
+                        neg_in=participants.neg_in)
+    neg_out_meta_review = meta_review_participants.objects.get(
+                        neg_out=participants.neg_out)
+    if pos_in_meta_review > 0 or pos_out_meta_review > 0 or
```
li-boxuan: See https://stackoverflow.com/questions/181530/styling-multi-line-conditions-in-if-statements

gamification/process/update.py
```diff
@@ -135,6 +136,29 @@ def get_participant_objects():
...
+def update_participants_data_with_meta_review(participants):
+    """
+    Update participants based on meta-review
+
+    This method updates every participant based on the meta-review
+    recieved or given. If the meta-review is completed then it calls
+    'Completed Meta-Review' activity.
+    """
+    pos_in_meta_review = meta_review_participants.objects.get(
+                        pos_in=participants.pos_in)
```
li-boxuan: This part makes me a bit confused. Does this ever work? Apparently `participants` come from `gamification.models`, which doesn't have `pos_in` field.

`meta_review_participants.objects.get(pos_in=some number)` doesn't make sense to me either. This returns an object whose `pos_in` field = `some number`.

.gitignore
```diff
@@ -488,3 +488,5 @@ _site/
 
 # Pytest profile files
 prof/
+*.orig
+.vscode/settings.json
```
li-boxuan: Create a new issue for changes you want to make that are unrelated to this pr.
Also, `.vscode` already exists in `.gitignore`.

gamification/process/update.py
sks444: One idea is that this method would take list of all the participants objects from the
meta-review partcipant model...and then iterate through each of the participant object from the list while filtering out only those participant whose usernames are in [active_newcomer_list](https://github.com/coala/community/blob/master/data/newcomers.py#L9)(as gamifiction system only work for those participant) and then if that participant have `completed meta-review` then get the object of that participant from the gamification participant model and add the meta-review activity for that participant.

sks444: I think maybe you're getting confused b/w `meta-review Participants` and `gamification Participants`, meta-review system works for all members at coala and gamification system works for only [active newcomers](https://github.com/coala/community/blob/master/data/newcomers.py#L9) at coala.

participants from `get_participant_objects()` are coming from gamification models, so they do not have fields like `pos_in` and `neg_out` etc..so you can't do[1](https://github.com/coala/community/pull/190/files#diff-49eb4e29b7797643ccd41983aeb4c575R148) as @li-boxuan mentioned there.

and where is `meta_review_completed` field? if you want to do something like `participant.meta_review_completed` with participant object you need a field of that name to the `Participant` model. I think you don't need to do something like that, when you find that a participant has completed meta-review, just add the `meta-review completed` activity for that participant by using [`add_activity`](https://github.com/coala/community/blob/master/gamification/models.py#L135) method.

Maybe rebase this commit on top of https://github.com/coala/community/pull/152 so that this pr passes CI and we can see you changes running.
> I'm testing these changes on another branch so that while pushing the changes, my PR doesn't gets bad with Travis test failing messages

.gitignore
shikharvaish28: Created https://github.com/coala/community/issues/194

.gitignore
```diff
@@ -488,3 +488,4 @@ _site/
 
 # Pytest profile files
 prof/
+*.orig
```
li-boxuan: Remove this

gamification/data/config.py
```diff
@@ -76,6 +76,9 @@ def create_badge_activities():
             name='Solved a area/genericbears bear proposal issue'),
         BadgeActivity(
             name='Solved a area/lintbears bear proposal issue'),
+        BadgeActivity(
+            name='Completed Meta-Review'
+        ),
```
li-boxuan: IMO it's better to keep the format consistent, i.e. put `),` at the end of last line

gamification/management/commands/update_participants_data.py
```diff
@@ -31,3 +33,5 @@ def handle(self, *args, **options):
             update_participants_data_with_issue(issue)
         for participant in get_participant_objects():
             award_badges(participant)
+        for participant in get_participant_objects_meta_review():
```
li-boxuan: IMO `get_meta_review_participant_objects` sounds better than `get_participant_objects_meta_review`

gamification/process/update.py
```diff
@@ -135,6 +136,34 @@ def get_participant_objects():
     return participants
 
 
+def get_participant_objects_meta_review():
+    meta_review_participant = meta_review_participants.objects.all()
+    return meta_review_participant
```
li-boxuan: Just simply do `return meta_review_participants.objects.all()` is enough

gamification/process/update.py
```diff
@@ -135,6 +136,34 @@ def get_participant_objects():
...
+                        pos_in=meta_review_participants.pos_in)
+    pos_out_meta_review = meta_review_participants.objects.get(
+                        pos_out=meta_review_participants.pos_out)
+    neg_in_meta_review = meta_review_participants.objects.get(
+                        neg_in=meta_review_participants.neg_in)
+    neg_out_meta_review = meta_review_participants.objects.get(
+                        neg_out=meta_review_participants.neg_out)
+    if pos_in_meta_review > 0 or pos_out_meta_review > 0 or \
+       neg_in_meta_review > 0 or neg_out_meta_review > 0:
+        participant.meta_review_completed = True
```
li-boxuan: Is this field defined anywhere? Moreover, IMO this field should belong to gamification participant, not meta-review participant. Seems you are passing a meta-review participant to this function `update_participants_data_with_meta_review`

gamification/process/update.py
```diff
@@ -135,6 +136,34 @@ def get_participant_objects():
...
+def update_participants_data_with_meta_review(participant):
+    """
+    Update participants based on meta-review
+
+    This method updates every participant based on the meta-review
+    received or given. If the meta-review is completed then it calls
+    'Completed Meta-Review' activity.
+    """
+    pos_in_meta_review = meta_review_participants.objects.get(
+                        pos_in=meta_review_participants.pos_in)
```
li-boxuan: I don't quite understand this. If you want to get the value of `pos_in` of the current meta-review participant (seems the parameter of this function is meta_review participant), you should do `pos_in_meta_review = participant.pos_in`

shikharvaish28:
I'm passing a meta-review participant from a list of participants extracted from ```get_meta_review_participant_objects```. Then, we iterate over that list and send each member to function ```update_participants_data_with_meta_review``` since the meta-review is done for each participant. If ```pos_in``` and other attributes are found to be positive then we call [add_activity](https://github.com/coala/community/blob/master/gamification/models.py#L135) to update the activity.

@li-boxuan @sks444 Your thoughts please.

meta_review/models.py
```diff
@@ -95,3 +95,12 @@ class Reaction(models.Model):
     giver = models.ForeignKey(Participant, related_name='give', null=True)
     receiver = models.ForeignKey(Participant, related_name='receive', null=True)
     review = models.ForeignKey(Comment, null=True)
+
+    def meta_review_add_activity(self):
```
li-boxuan: meta-review activity shouldn't be defined in the meta-review app, but in the gamification app.

shikharvaish28: Will change it. Apart from that is everything fine?

li-boxuan: I think this is the major problem. Once the logic gets correct then we can focus on minor issues like coding style, etc.

gamification/models.py
```diff
@@ -186,3 +187,13 @@ def add_badges(self, activities):
         badges = self.find_badges_for_activity(activities)
         for badge in badges:
             self.badges.add(badge)
+
+
+def meta_review_add_activity(self):
+    """
+    Add a meta-review activity to the participant
+    """
+    activity = Reaction.object.create(
```
li-boxuan: This part does not make sense to me. I don't understand what you are doing with `Reaction` here. Once you know if a participant has finished the meta-review process, the rest staff has nothing to do with the meta-review app.

li-boxuan: What you said at https://github.com/coala/community/pull/190#issuecomment-423495158 looks good. However, you are not handling `update the activity` correctly.

shikharvaish28: > This part does not make sense to me. I don't understand what you are doing with `Reaction` here. Once you know if a participant has finished the meta-review process, the rest staff has nothing to do with the meta-review app.

In ```meta_review_add_activity ```, I'm using the details from```Reaction``` class of ```meta-review``` app to add the ```Completed Meta-Review``` activity (In the same way as that of ```add_activity``` in the ```gamification```.



> What you said at [#190 (comment)](https://github.com/coala/community/pull/190#issuecomment-423495158) looks good. However, you are not handling `update the activity` correctly here.
> 
> Moreover, this should not be a standalone function. At least make it a method of `Participant` class in `gamification/models.py`. Maybe we don't even need it - you can simply put the logic in `update_participants_data_with_meta_review` function. After you implement the correct logic, we can start talking about coding style.

I added this as a function because of the pattern followed by ```add_activity``` of ```gamification```.

If you feel that I should remove this function as a whole and implement the logic in ```update_participants_data_with_meta_review```, I can do it.

li-boxuan: > I'm using the details fromReaction class of meta-review app to add the Completed Meta-Review activity

An activity should be an instance of `Activity` class. When you do `Reaction.object.create` you are creating an instance of `Reaction`, which does not make sense (and `Reaction` does not have `activity` field btw).

> I added this as a function because of the pattern followed by add_activity of gamification.

Okay, I didn't notice that point ;) Then don't put it in `models.py`. But we can put this thing aside and make the logic go correctly first.

shikharvaish28:
> https://gci-leaders.netlify.com/data.json

The link is working and there's no issue on my local machine too while running ```coala``` .  Why is it failing here?

gamification/process/update.py
```diff
@@ -135,6 +136,27 @@ def get_participant_objects():
...
+    received or given. If the meta-review is completed then it calls
+    'Completed Meta-Review' activity.
+    """
+    pos_in_meta_review = participant.pos_in
+    pos_out_meta_review = participant.pos_out
+    neg_in_meta_review = participant.neg_in
+    neg_out_meta_review = participant.neg_out
+    if pos_in_meta_review > 0 or pos_out_meta_review > 0 or \
+       neg_in_meta_review > 0 or neg_out_meta_review > 0:
+        participant.meta_review_add_activity(participant)
```
li-boxuan: I don't think this would work as participant does not have `meta_review_add_activity` method.

shikharvaish28:
By mistake, I clicked on ```Update``` and it caused the issue but it's now fixed.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.linting.enabled": false
+}
```
sks444: Please do not push this file.

gamification/process/update.py
```diff
@@ -135,6 +136,33 @@ def get_participant_objects():
...
+
+
+def update_participants_data_with_meta_review(participant):
+    """
+    Update participants based on meta-review
+
+    This method updates every participant based on the meta-review
+    received or given. If the meta-review is completed then it calls
+    'Completed Meta-Review' activity.
+    """
```
sks444: First check if this participant is in the [active_newcomer_list](https://github.com/coala/community/blob/da5ada362c8fc3ae2cb179739a4babbfad1e7d1f/data/newcomers.py#L9), as gamification only work for those participant.
```
if participant.login in active_newcomers:
     ....
```

gamification/process/update.py
```diff
@@ -135,6 +136,33 @@ def get_participant_objects():
...
+    neg_out_meta_review = participant.neg_out
+    if pos_in_meta_review > 0 or pos_out_meta_review > 0 or \
+       neg_in_meta_review > 0 or neg_out_meta_review > 0:
+        """
+        Add a meta-review activity to the participant
+        """
+        activity = Activity.object.create(
+            name=participant,
+            activity='Completed Meta-Review')
+        participant.activities.add(activity)
```
sks444: Now, this participant is a meta-review object.
You have to get a gamification participant first.
```
from gamification.models import Participant
gam_participant = Participant.objects.get_or_create(username=participant.login)
```

gamification/process/update.py
```diff
@@ -135,6 +136,33 @@ def get_participant_objects():
...
+    pos_in_meta_review = participant.pos_in
+    pos_out_meta_review = participant.pos_out
+    neg_in_meta_review = participant.neg_in
+    neg_out_meta_review = participant.neg_out
+    if pos_in_meta_review > 0 or pos_out_meta_review > 0 or \
+       neg_in_meta_review > 0 or neg_out_meta_review > 0:
+        """
+        Add a meta-review activity to the participant
+        """
+        activity = Activity.object.create(
```
sks444: Try to use [add_activity](https://github.com/coala/community/blob/da5ada362c8fc3ae2cb179739a4babbfad1e7d1f/gamification/models.py#L135) method.

gamification/process/update.py
shikharvaish28: Thanks!

gamification/process/update.py
```diff
@@ -135,6 +136,33 @@ def get_participant_objects():
...
+
+    This method updates every participant based on the meta-review
+    received or given. If the meta-review is completed then it calls
+    'Completed Meta-Review' activity.
+    """
+    pos_in_meta_review = participant.pos_in
+    pos_out_meta_review = participant.pos_out
+    neg_in_meta_review = participant.neg_in
+    neg_out_meta_review = participant.neg_out
+    if pos_in_meta_review > 0 or pos_out_meta_review > 0 or \
```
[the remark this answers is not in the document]
li-boxuan: Avoid using `\`. If conditions are too long, use `(`, `)`.
```python
if (xxx or xxx or
        xxx or xxx):
```

li-boxuan:
It would be much better if you can add some tests, e.g. creating two meta-review participants, one of whom has done a meta-review while the other one hasn't. Then obviously the former one completes the meta-review activity while the latter one doesn't complete.

gamification/process/update.py
shikharvaish28: While adding the tests, I'm receiving this traceback 
``` 
    def update_participants_data_with_meta_review(meta_review_participant):
        """
        Update participants based on meta-review

        This method updates every participant based on the meta-review
        received or given. If the meta-review is completed then it calls
        'Did a meta-review or received a meta-review' activity.
        """
>       if meta_review_participant.login in active_newcomers:
E       AttributeError: 'Participant' object has no attribute 'login'

gamification/process/update.py:152: AttributeError
```

@sks444

li-boxuan: @shikharvaish28 It's hard to tell the reason unless you provide all code, but I guess it's because the `meta_review_participant` you are using here is NOT a meta-review participant, but a gamification participant, which does not have attribute `login`.

shikharvaish28: @li-boxuan @sks444 Please have a look!

gamification/process/update.py.orig
```diff
@@ -0,0 +1,172 @@
+import logging
```
li-boxuan: Remove this file.

gamification/tests/test_models.py
```diff
@@ -341,3 +345,18 @@ def test_add_badges_method(self):
 
         # After applying add_badge method
         self.assertEquals(test_participant.badges.count(), 0)
+
+    def test_update_participants_data_with_meta_review(self):
+        # Test the participant who has completed the meta-review activity
+        Participant.objects.create(username='shikharvaish28')
+        test_meta_review_participant = Participant.objects.get(
+                                       username='shikharvaish28')
```
li-boxuan: This is wrong. You are mixing up gamification participants and meta-review participants. You also need to import `Participant` from `meta-review` app.

gamification/tests/test_models.py
```diff
@@ -341,3 +347,18 @@ def test_add_badges_method(self):
 
         # After applying add_badge method
         self.assertEquals(test_participant.badges.count(), 0)
+
+    def test_update_participants_data_with_meta_review(self):
+        # Test the participant who has completed the meta-review activity
+        meta_review_participant.objects.create(login='test2')
+        test_meta_review_participant = meta_review_participant.objects.get(
```
shikharvaish28: I have an issue here. Here's what I try to do:

1. Create two meta-review participants. One with meta-review completed while the other has not.
2. Pass both of them through ```update_participants_data_with_meta_review()``` and update.
3. The one with meta-review completed would have added one activity unlike the other.


Now the issue is that when I'm creating the user through ```meta_review_participant```, I'm unable to get the activities count since it's a part of the ```gamification```.

li-boxuan: Besides creating a meta-review participant, you also need to create a **corresponding** gamification participant.

shikharvaish28: I tried to create a ```gamification``` participant with the same name as that of ```meta-review``` one. Now, the problem is how to link them. 
Though the activity is being added in ```meta-review```, how can I update it in ```gamification``` because in gamification, it's still showing activity count as 0 which in ideal situations should be 0?

gamification/process/update.py
```diff
@@ -135,6 +137,37 @@ def get_participant_objects():
...
+            meta_review_participant.neg_in > 0 or
+                meta_review_participant.neg_out > 0):
+            """
+            Add a meta-review activity to the participant
+            """
+            name = meta_review_participant
+            activity = 'Did a meta-review or received a meta-review'
+            performed_at = Activity.performed_at
+            updated_at = Activity.updated_at
+            Participant.add_activity(name,
```
li-boxuan: Seems this is wrong. You should find the corresponding `gamification participant`. And do sth like `gamification_participant.add_activity()` instead of `Participant.add_activity`.

gamification/tests/test_models.py
li-boxuan: > Though the activity is being added in meta-review

You only need to play around with meta-review app to fetch some data you need. All the other stuff are related to `gamification`. `Activity` has nothing to do with `meta-review participant`.

gamification/process/update.py
```diff
@@ -135,6 +137,39 @@ def get_participant_objects():
...
+
+def update_participants_data_with_meta_review(meta_review_participant):
+    """
+    Update participants based on meta-review
+
+    This method updates every participant based on the meta-review
+    received or given. If the meta-review is completed then it calls
+    'Did a meta-review or received a meta-review' activity.
+    """
+    if meta_review_participant.login == active_newcomers:
```
jayvdb: the tests say that line 153 - 166 here are not being tested at all, which means in the tests, this is always False. 

```console
gamification/process/update.py               69      8     19      1    85%   153-166, 152->153
```

And that is because this line is ridiculously wrong.  print out each side of this `==` and study the printout.

li-boxuan: Add print function, and use `.ci/build.sh` to see the output

shikharvaish28: Tried the same but I'm receiving this traceback:

```
(test_env) shikharvaish@Shikhars-MacBook-Air /Coding/GSoC/community (meta-review)
$ .ci/build.sh
+ rm db.sqlite3
rm: db.sqlite3: No such file or directory
+ true
+ rm -rf private _site public
+ mkdir private _site public
+ META_REVIEW_DATA=meta_review.json
+ EXPORTED_DATA='static/tasks.yaml static/instances.yaml static/meta_review.json'
+ ISSUES_JSON=issues.json
+ python manage.py fetch_deployed_data --allow-failure _site static/tasks.yaml static/instances.yaml static/meta_review.json
Traceback (most recent call last):
  File "manage.py", line 24, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/shikharvaish/anaconda3/envs/test_env/lib/python3.6/site-packages/django/core/management/__init__.py", line 367, in execute_from_command_line
    utility.execute()
  File "/Users/shikharvaish/anaconda3/envs/test_env/lib/python3.6/site-packages/django/core/management/__init__.py", line 359, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/shikharvaish/anaconda3/envs/test_env/lib/python3.6/site-packages/django/core/management/base.py", line 294, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/shikharvaish/anaconda3/envs/test_env/lib/python3.6/site-packages/django/core/management/base.py", line 345, in execute
    output = self.handle(*args, **options)
  File "/Coding/GSoC/community/data/management/commands/fetch_deployed_data.py", line 35, in handle
    upstream_deploy_url = get_upstream_deploy_url(repo_name, hoster)
  File "/Coding/GSoC/community/community/git.py", line 191, in get_upstream_deploy_url
    repo = get_upstream_repo()
  File "/Coding/GSoC/community/community/git.py", line 153, in get_upstream_repo
    parent = get_parent_repo(remote)
  File "/Coding/GSoC/community/community/git.py", line 119, in get_parent_repo
    repo = get_irepo(url)
  File "/Coding/GSoC/community/community/git.py", line 108, in get_irepo
    hoster = get_ihoster(url)
  File "/Coding/GSoC/community/community/git.py", line 87, in get_ihoster
    if url.resource == 'github.com':
AttributeError: 'list' object has no attribute 'resource'
```

gamification/process/update.py
```diff
@@ -135,6 +137,42 @@ def get_participant_objects():
...
+            meta_review_participant.neg_in > 0 or
+                meta_review_participant.neg_out > 0):
+            """
+            Add a meta-review activity to the participant
+            """
+            activity = 'Did a meta-review or received a meta-review'
+            created_at = reaction_name.created_at
+            updated_at = '2017-08-24 05:59:31+00:00'
+            gamification_participant.add_activity(
+                                                0,
```
jayvdb: horrible indentation. find a nicer way.

gamification/process/update.py
```diff
@@ -135,6 +137,42 @@ def get_participant_objects():
...
+    in the active newcomers list then it will check if the meta-review
+    is complete or not and update the activity accordingly.
+    """
+    active_newcomers_list = active_newcomers()
+    if meta_review_participant.login in active_newcomers_list:
+        gamification_participant = Participant.objects.get(
+            username=meta_review_participant.login)
+        reaction_name = Reaction.objects.get(id=meta_review_participant.login)
+        if (meta_review_participant.pos_in > 0 or
+            meta_review_participant.pos_out > 0 or
```
jayvdb: this violates pep8

gamification/process/update.py
```diff
@@ -135,6 +137,38 @@ def get_participant_objects():
...
+                meta_review_participant.neg_in > 0 or
+                meta_review_participant.neg_out > 0):
+            """
+            Add a meta-review activity to the participant
+            """
+            activity = 'Did a meta-review or received a meta-review'
+            created_at = reaction_name.created_at
+            updated_at = None
+            gamification_participant.add_activity(0, activity,
+                                                  created_at, updated_at,)
```
jayvdb: a `,` before a `)` is completely useless.

the concept is called trailing comma , because it adds a comma and the end of the line -- the absolute end of the line, with nothing after it.

but also, use hanging indent for these args, dont put the args on the same line as the function name.  that isnt nice when the function name is long, as the args need to be squished on the RHS of the page.

gamification/process/update.py
```diff
@@ -135,6 +137,38 @@ def get_participant_objects():
...
+        gamification_participant = Participant.objects.get(
+            username=meta_review_participant.login)
+        reaction_name = Reaction.objects.get(id=meta_review_participant.login)
+        if (meta_review_participant.pos_in > 0 or
+                meta_review_participant.pos_out > 0 or
+                meta_review_participant.neg_in > 0 or
+                meta_review_participant.neg_out > 0):
+            """
+            Add a meta-review activity to the participant
+            """
```
li-boxuan: Use `#` for comment here since it's single line anyway.

gamification/process/update.py
```diff
@@ -135,6 +137,38 @@ def get_participant_objects():
...
+            activity = 'Did a meta-review or received a meta-review'
+            created_at = reaction_name.created_at
+            updated_at = None
+            gamification_participant.add_activity(0, activity, created_at,
+                                                  updated_at)
```
li-boxuan: I am not 100% sure here, but I suspect this would add repetitive activities to newcomers, because you don't check whether the newcomer has already completed a meta-review activity.

li-boxuan: Also add a testcase to verify that ;)
especially if you believe I am wrong here.

gamification/tests/test_models.py
```diff
@@ -341,3 +349,48 @@ def test_add_badges_method(self):
 
         # After applying add_badge method
         self.assertEquals(test_participant.badges.count(), 0)
+
+    def test_update_participants_data_with_meta_review(self):
```
li-boxuan: Can you split this long test case to several smaller test cases?

shikharvaish28: Can you please throw some light on how you want me to split it? It's because all of the 3 participants here are for the same stuff so it's becoming difficult to identify which form do I split it in. @li-boxuan

gamification/process/update.py
shikharvaish28: > I suspect this would add repetitive activities to newcomers

Sounds logical but in order to do this, I will have to add in a flag to ensure if the meta-review has been completed. 
Or should I create another list for the participants who have completed the meta-review. While passing on a participant's name, we can check if he's in the list or not. In case it is, we can break the statement.

li-boxuan: Maybe before adding a new activity, check if the user has already finished a meta-review activity. I’m not sure how gamification app avoids other duplicated activities, maybe you can learn from them.

gamification/process/update.py
```diff
@@ -135,6 +137,42 @@ def get_participant_objects():
...
+        if (meta_review_participant.pos_in > 0 or
+                meta_review_participant.pos_out > 0 or
+                meta_review_participant.neg_in > 0 or
+                meta_review_participant.neg_out > 0):
+            # Add a meta-review activity to the participant
+            activity = 'Did a meta-review or received a meta-review'
+            created_at = reaction_name.created_at
+            updated_at = None
+            gamification_participant.add_activity(0, activity, created_at,
+                                                  updated_at)
```
sks444: Maybe add an `info level` log here saying that `this participant(get the username of the participant with gamification_participant.username ) has performed meta-review activity`.

gamification/process/update.py
```diff
@@ -135,6 +137,42 @@ def get_participant_objects():
...
+    in the active newcomers list then it will check if the meta-review
+    is complete or not and update the activity accordingly.
+    """
+    active_newcomers_list = active_newcomers()
+    if (meta_review_participant.login in active_newcomers_list and
+            meta_review_participant.login not in meta_review_completed_list):
+        # Get the corresponding gamification participant
+        gamification_participant = Participant.objects.get(
+            username=meta_review_participant.login)
+        reaction_name = Reaction.objects.get(id=meta_review_participant.login)
```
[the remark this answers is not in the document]
shikharvaish28: ```Reaction.id``` and ```meta_review_participant.login```  are the name of the participant. (I've cross checked it by printing both of them.

gamification/process/update.py
```diff
@@ -135,6 +137,38 @@ def get_participant_objects():
...
+def get_meta_review_participant_objects():
+    return meta_review_participants.objects.all()
+
+
+def update_participants_data_with_meta_review(meta_review_participant):
+    """
+    Update participants based on meta-review
+
+    This method updates every participant based on the meta-review
+    received or given. If the gamification participant is
```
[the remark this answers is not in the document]
shikharvaish28: I've done this already [here](https://github.com/coala/community/pull/190/files#diff-49eb4e29b7797643ccd41983aeb4c575R157) :)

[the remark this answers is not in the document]
sks444: No, I meant the sentence should be `If the meta-review participant is in..`?

gamification/process/update.py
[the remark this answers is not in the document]
li-boxuan: Really? How did you cross check that? That would be super weird.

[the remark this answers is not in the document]
shikharvaish28: In ```test_model.py```, I printed both of them separately and for seeing the output, ran ```pytest -s```

li-boxuan:
I cannot see anyone who has finished `meta-review` activity on https://deploy-preview-190--coala-community.netlify.com/gamification/

shikharvaish28:
> I cannot see anyone who has finished `meta-review` activity on https://deploy-preview-190--coala-community.netlify.com/gamification/

I'm myself able to understand why is it happening because the tests are running perfectly fine.

gamification/process/update.py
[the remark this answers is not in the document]
shikharvaish28: The ```meta_review_participant.login``` [output](https://pastebin.com/LEGhK3tS) and ```reaction``` [output](https://pastebin.com/6X6y5Kfx). 

The ```.ci/build.sh``` [output](https://pastebin.com/J9m8uBnt)

gamification/tests/test_models.py
```diff
@@ -341,3 +349,64 @@ def test_add_badges_method(self):
...
         self.assertEquals(test_participant.badges.count(), 0)
+
+    def test_update_participants_data_with_meta_review_complete(self):
+        # Test the participant who has completed the meta-review activity
+        active_newcomers_list = active_newcomers()
+        active_newcomers_list.extend(['test2', 'test3'])
+        # meta-review participant
+        meta_review_participant.objects.create(login='test2')
+        test_meta_review_participant = meta_review_participant.objects.get(
+                                       login='test2')
```
li-boxuan: Unnecessary. `meta_review_participant.objects.create(login='test2')` will return the instance.

gamification/tests/test_models.py
```diff
@@ -341,3 +349,64 @@ def test_add_badges_method(self):
...
+
+        Reaction.objects.create(id='test2')
```
li-boxuan: Well, this is definitely inappropriate and prone to error. You are creating a reaction with id also equals to `test2`, but the participant you created is also with id `test2`.

gamification/process/update.py
```diff
@@ -135,6 +137,45 @@ def get_participant_objects():
...
+    in the active newcomers list then it will check if the meta-review
+    is complete or not and update the activity accordingly.
+    """
+    active_newcomers_list = active_newcomers()
+    if (meta_review_participant.login in active_newcomers_list and
+            meta_review_participant.login not in meta_review_completed_list):
+        # Get the corresponding gamification participant
+        gamification_participant = Participant.objects.get(
+            username=meta_review_participant.login)
+        reaction = Reaction.objects.get(id=meta_review_participant.login)
```
li-boxuan: This is definitely wrong. I checked the output you print out https://pastebin.com/6X6y5Kfx. You have the illusion that this works, because your test is wrong. In your test, both participant and reaction share the same id (primary_key), which gives you the wrong belief that you are correct.

That's why I said you should run `.ci/build.sh` and show the log to prove this works. You can use that for debugging, but you don't need to prove anymore as I am pretty confident now this is wrong.

li-boxuan: First you need to fix the test case. See https://github.com/coala/community/pull/190#discussion_r229155532 so that your test won't pass (which is good coz now you know your code is wrong).

Then read the definition of `Reaction` model. Understand the relationships between those models.

shikharvaish28: In order to test things, I used my user handle ```shikharvaish28```. Ideally, I should be able to get  ```reaction``` using ```id``` which is the primary key of ```Reaction``` model.

```python
test_meta_review_participant = meta_review_participant.objects.get(login='shikharvaish28')
reaction = Reaction.objects.get(id=test_meta_review_participant.login)
``` 
Instead, I am getting ```DoesNotExist: Reaction matching query does not exist.``` Can you please tell me why is it so?

li-boxuan: > Since login is the primary key of Participant, I am expecting it to be linked in someway to id of Reaction.

Your assumption is wrong. Django does not work like that.

I strongly recommend you to read https://docs.djangoproject.com/en/2.1/topics/db/queries/, esepecially understand how Foreign Key works.

shikharvaish28:
@li-boxuan I'm closing this PR since I'm unable to solve this. I would prefer to do some other issues as of now. Kindly unassign me from this one so that this is free for someone else to take up.

PR #210, opened 2018-10-16
DNM: GCI 2017 issue data

jayvdb:
Adds issues.yaml used by https://github.com/coala/community/pull/206

Adds example issues.yaml to be used by https://github.com/coala/community/pull/206

Not to be merged, as this is too coala specific.

We need to find an alternative storage location for this.


jayvdb:
No longer needed

PR #211, opened 2018-10-16
requirements.txt: Bump IGitt dependency

jayvdb:
Use latest IGitt, and also include a commit which
unpins package `cryptography` so it is easier to
install IGitt, otherwise the old version required of
`cryptography` may require the user has a compiler
set up, as wheels may not work on their system.

### DIFF
--- a/.moban.yaml
+++ b/.moban.yaml
@@ -21,7 +21,7 @@ dependencies:
   - django
   - django-distill
   - django-eventtools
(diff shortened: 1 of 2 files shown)


jayvdb:
This is blocked by https://gitlab.com/gitmate/open-source/IGitt/issues/114 , or some other solution needed.

PR #215, opened 2018-10-23
GCI: Update tasks

jayvdb:
A rather large sprawling set of changes.


KVGarg:
@jayvdb The `pytest` is raising error because some files can't be imported in which two are the files that you have created - `update_issue_tasks.py` and `add_gci_label.py`. Can you take a look at it by taking out some time ;-)
Also need re-base.

PR #217, opened 2018-10-29
.moban.yaml: Disable openhub module

sks444:
openhub app is broken at the moment which
is causing netlify to fail. So, this commit
disables the openhub module until we find out
the problem and fix it.


jayvdb:
ack https://github.com/coala/community/pull/217/commits/40109d05d0d88ad9c217a6baa2c294add436a0e2

jayvdb:
@gitmate-bot ff

Issue #201, opened 2018-10-13
OpenHub without token generates empty page


jayvdb:
If no OH_TOKEN was given during the build, the openhub/ page will be empty.

The `openhub/` page should explain how to enable the app, or the top level index page could be dynamic and place the `openhub/` link in a 'Disabled section' with a note about using tokens.

bhawesh96:
I tried to reproduce the issue and went through the related to code.
Will it be a feasible fix if I call ```CurrentUserStore.fetchUser()``` upon clicking 'Save'. If returns ```info```, then allow Save, else, show error message in the login dialog itself?

PR #219, opened 2018-11-02
OpenHub: /openhub shows how to get OH_TOKEN if not set

kx-chen:
/openhub now shows a short message on how to get an OH_TOKEN if it's not set

Closes https://github.com/coala/community/issues/201

Unsure if having the message in the code like that is okay, but I was unable to think of another way to do it.

### DIFF
--- a/openhub/views.py
+++ b/openhub/views.py
@@ -1,9 +1,20 @@
 from django.shortcuts import render
 
 from openhub.models import PortfolioProject
+from openhub.oh_token import OH_TOKEN
 
 
 def index(request):
-    projects = PortfolioProject.objects.all()
-    args = {'projects': projects}
+    errors = []
+    projects = []
+    if OH_TOKEN is None:
+        errors.append("""
+        OH_TOKEN is not specified as an environment variable. <br/>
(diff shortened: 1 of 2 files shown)


templates/openhub.html
```diff
@@ -10,6 +10,7 @@
   </head>
   <body>
     <h1>All of our Portfolio Projects</h1>
+    <p>{{ error|safe }}</p>
```
jayvdb: Spaces around `|`

openhub/views.py
```diff
@@ -1,9 +1,20 @@
...
 
 from openhub.models import PortfolioProject
+from openhub.oh_token import OH_TOKEN
 
 
 def index(request):
-    projects = PortfolioProject.objects.all()
-    args = {'projects': projects}
+    error = ''
+    projects = ''
```
jayvdb: Should be array

jayvdb:
ack https://github.com/coala/community/pull/219/commits/f294015547ec2c0e5844f6e44ed68e5fdbadb252

jayvdb:
@gitmate-bot ff

PR #236, opened 2019-02-09
Add an attribute to indicate task state

### LINKED ISSUE #66: Add status to issues tasks in tasks.yaml
When a task has `max_instances: 1` and an issues URL , it can be annotated with an extra attribute to indicate the task status: available, claimed or completed.

Instance are exported as instances.yaml , and indicate whether the task is in progress or completed.

Completed issues are the easiest to work with.

In progress tasks are slightly more complicated because the publish instances are filtered to remove tasks still in early phases of the task lifecycle.  See https://github.com/coala/community/blob/master/gci/students.py#L84

KVGarg:
This commit adds an extra attribute to indicate the task state:AVAILABLE, 
CLAIMED or COMPLETED to those tasks which have `max_instances=1` 
and an issues URL. Based upon these COMPLETED tasks, the instances
are being filtered out during cleansing.

Closes #66


gci/task.py
```diff
@@ -81,12 +83,44 @@ def beginner_tasks(tasks):
 
 def strip_mentors(tasks):
     for task in tasks.values():
-        del task['mentors']
+        if task.__contains__('mentors'):
+            del task['mentors']
+
 
+def cleanse_tasks(tasks, tokens):
+
```
areebbeigh:
```suggestion
```

gci/task.py
```diff
@@ -81,12 +83,44 @@ def beginner_tasks(tasks):
...
-def cleanse_tasks(tasks):
     cleansed_tasks = published_tasks(tasks)
 
     strip_mentors(tasks)
 
     return cleansed_tasks
+
+
+def get_task_state(task_url, tokens):
+
```
areebbeigh:
```suggestion
```

gci/task.py
```diff
@@ -81,12 +84,49 @@ def beginner_tasks(tasks):
...
+            del task['mentors']
 
 
 def cleanse_tasks(tasks):
+    tokens = {
+        'GH_TOKEN': os.environ.get('GH_TOKEN'),
+    }
+
+    for task in tasks.values():
+        if task['max_instances'] == 1 \
```
frextrite: Isn't the explicit use of line continuation discouraged?

KVGarg: yes, will make it correct :+1:

gci/students.py
```diff
@@ -1,8 +1,7 @@
-import re
 import logging
+import re
```
li-boxuan: Reason for this change?

KVGarg: It was a result of auto-formatting by pyCharm IDE. Will undo that change

KVGarg:
@jayvdb In the travis CI logs (on-wards [line 943](https://travis-ci.org/coala/community/jobs/499891113#L943)), a GITHUB API call limit reached is being highlighted. Do we need to take care about this ?

Issue #243, opened 2019-05-08
Build failing due to latest version of django getting installed


KVGarg:
Django v2.1.7 should be installed as it depends upon Ubuntu Trusty v14.04 and makes the CI builds green. Installing latest version is leading to problems as it requires Sqlite v3.8.3 or later version which isn't available in Ubuntu Trusty. Available for now in Ubuntu Xenial image.

jayvdb:
The solution is to use the xenial image.  Then you saw other problems with Django 2.2.  They should be documented here in this issue.

And there were problems also with Django 2.1 ?  They should be documented in a separate issue.

KVGarg:
> xenial image. 
Using xenial image is installing later versions of py>1.5.2, pluggy>0.6.0. One of the reasons due to which build is failing when using xenial. Ubuntu xenial image has to be used if want to work with django v2.2 as it no longer supports sqlite v3.8.2

When using xenial image- 
Issues with Django v2.1.7 when pytest v4.4.0 or later is installed with it? Please refer [build](https://travis-ci.org/coala/community/jobs/532912493) 
Issues with Django v2.1.7 when pytest v3.6.4? Please refer [build](https://travis-ci.org/coala/community/jobs/532920800)
Issues with Django v2.2.1 when pytest v4.4.2? Please refer [build](https://travis-ci.org/KVGarg/community/jobs/532835986)

Issue #241, opened 2019-02-14
moban sync needed


KVGarg:
See [moban](https://gitlab.com/coala/mobans) readme for steps

A Newcomer issue

PR #244, opened 2019-05-08
Install Django v2.1 and Sync moban

KVGarg:
The CI builds were failing due to un-synced moban &
latest versions installed of Django and pytest-xdist.
The django v2.1.8 or later version doesn't support
sqlite v3.8.2 anymore which is a pre-installed package
in Ubuntu Trusty. And, latest pytest-xdist needs
pytest v4.4.0 but v3.6.4 is being installed.

Closes https://github.com/coala/community/issues/243
Closes https://github.com/coala/community/issues/241, https://github.com/coala/community/pull/246

### DIFF
--- a/.moban.yaml
+++ b/.moban.yaml
@@ -18,7 +18,7 @@ packages:
 dependencies:
   - git+https://gitlab.com/coala/coala-utils.git
   - git-url-parse
-  - django
+  - django>2.1,<2.2
   - django-distill
   - django-eventtools
   - git+https://gitlab.com/gitmate/open-source/IGitt.git@1fa5a0a21ea4fb8739d467c06972f748717adbdc
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -12,7 +12,7 @@ coverage-config-reload-plugin~=0.2
 codecov~=2.0.5
 moban~=0.3.3 ; python_version > '3.0'
 packaging~=16.8
-pytest~=3.6.1
+pytest~=3.6.4
 pytest-cov~=2.4
 pytest-django~=3.3.3
 pytest-env~=0.6.0
@@ -24,7 +24,6 @@ pytest-reorder~=0.1.0
 git+https://github.com/jayvdb/pytest-reqs@coala#egg=pytest-reqs
 pytest-timeout~=1.3.0
 pytest-travis-fold~=1.3.0
-pytest-xdist~=1.15
 requests-mock~=1.2
 ipdb~=0.11
 pip<10
--- a/.gitignore
+++ b/.gitignore
@@ -87,6 +87,13 @@ ipython_config.py
 # pyenv
 .python-version
 
+# pipenv
+#   According to pypa/pipenv#598, it is recommended to include Pipfile.lock in version control.
+#   However, in case of collaboration, if having platform-specific dependencies or dependencies
+#   having no cross-platform support, pipenv may install dependencies that don't work, or not
+#   install all needed dependencies.
+#Pipfile.lock
+
 # celery beat schedule file
 celerybeat-schedule
 
@@ -152,6 +159,7 @@ pip-selfcheck.json
 # Windows rules
 # Windows thumbnail cache files
 Thumbs.db
+Thumbs.db:encryptable
 ehthumbs.db
 ehthumbs_vista.db
 
@@ -307,6 +315,8 @@ tags
 # .idea/modules.xml
 # .idea/*.iml
 # .idea/modules
+# *.iml
+# *.ipr
 
 # CMake
 cmake-build-*/
@@ -419,8 +429,14 @@ DerivedData/
 *.perspectivev3
 !default.perspectivev3
 
-# Eclipse rules
+## Xcode Patch
+*.xcodeproj/*
+!*.xcodeproj/project.pbxproj
+!*.xcodeproj/xcshareddata/
+!*.xcworkspace/contents.xcworkspacedata
+/*.gcno
 
+# Eclipse rules
 .metadata
 bin/
 tmp/
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 git+https://gitlab.com/coala/coala-utils.git
 git-url-parse
-django
+django>2.1,<2.2
 django-distill
 django-eventtools
 git+https://gitlab.com/gitmate/open-source/IGitt.git@1fa5a0a21ea4fb8739d467c06972f748717adbdc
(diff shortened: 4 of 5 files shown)


KVGarg:
@jayvdb The Travis CI build is failing as your forked [pytest-reqs coala branch](https://github.com/jayvdb/pytest-reqs/tree/coala) is behind from source repository. Can you please re-base it, to get the build green?
Also, `netlify` build is also failing due to usage of old version of ubuntu trusty version. Netlify has released `xenial` usage. Can you please change the build settings for this repository in `settings -> Build & Deploy -> Continuous Deployment -> Build image selection`

KVGarg:
Reverting pytest-xdist back to `~=1.15` as coala-moban now support pytest min. version 4.4.0 :+1:

KVGarg:
@sks444 Doing moban sync, is changing `pytest version` to `v4.4.0` which is causing Travis CI build failure. We have to set `pytest v3.6.1` and `pytest-xdist v1.26.1` to make Travis CI green otherwise it will lead to Build failure

.moban.yaml
```diff
@@ -18,7 +18,7 @@ packages:
 dependencies:
   - git+https://gitlab.com/coala/coala-utils.git
   - git-url-parse
-  - django
+  - django==2.1.7
```
jayvdb: The "problem" is django 2.2 , so it should be `django>2.1,<2.2`

It would also be acceptable to not touch this at all, allowing any version of Django including 2.2, and use xenial image in our CI.

The problem is our CI, not Django.

test-requirements.txt
```diff
@@ -12,7 +12,7 @@ coverage-config-reload-plugin~=0.2
 codecov~=2.0.5
 moban~=0.3.3 ; python_version > '3.0'
 packaging~=16.8
-pytest~=3.6.1
+pytest~=4.4.0
```
jayvdb: this needs to be the same as the version specified by coala mobans , which should be ok to increase because I've fixed the coverage plugins

KVGarg: Yes! It was generated by moban sync. Doing suggested changes will make it same as specified by coala mobans 😇

.moban.yaml
KVGarg: Yes! I also figures out first that the problem was due to different ubuntu version ie. trusty which was having sqlite v3.8.2 installed and doesn’t support later version of sqlite which was a prerequisite of django 2.2.1 which was getting installed from `requirements.txt`. Then later on i checked that django v2.2.1 was getting installed which i wasn’t familiar with and tried to roll back to v2.1.7 at which our builds were green. 
So, we can change to ubuntu xenial image by specifying `dist: xenial` in travis.yaml, but on netlify when I’ve changes my deployed fork repo to run on ubuntu xenial it wasn’t getting executed. The build was getting failed due to an unkonown error. Can you please check into it by changing settings for coala-communtiy repo on netlify. If the build gets failed then I guess we have to stick with ubuntu trusty version with django v2.1.x

test-requirements.txt
jayvdb: Yes, but then you are also proposing to change the version in the template at the mobans repo, which would mean it needs to change here again.  We need one version everywhere, so we need to find one set of versions which work everywhere.

KVGarg:
| Build Number | Versions info                                               |
| ------------------| ------------------------------------------------------|
| [1739](https://travis-ci.org/coala/community/jobs/532912493)  | Django v2.1.7 and pytest v4.4.2 |
| [1741](https://travis-ci.org/coala/community/jobs/532920800)  | Django v2.1.7 and pytest v3.6.4 |

@jayvdb Above is the build log with mentioned versions installed. And the error that is being raised during requirements check has been solved recently https://github.com/di/pytest-reqs/. This is getting installed from your fork which is behind from original one. Rebasing will make it grren if we go with 2nd possibilty

KVGarg:
Can’t we modify our tests so that it gets passed for latest version of pytest v4.4.0?

KVGarg:
The netlify build is taking longer than 15 minutes which is raising timeout error. The build should finish within 15 mins to get successful green! We need to find a way such that it takes less time. 
We can possibly reduce time by loading the last 204 tasks that are being loaded for which an api call is being made to get results. Each time a build run it makes those 204 api calls. Storing them somewhere will definitely reduce time

KVGarg:
For now the netlify has got passed, but still the build time is an issue or say warning. We have to take care about this.

jayvdb:
This PR uses https://github.com/coala/community/pull/244/commits/4534ac9cf12d70fcccfd15d46f481a92185b68a1 ; it needs to be based on the exact same commit as found in https://github.com/coala/community/pull/246 , which is currently https://github.com/coala/community/commit/9d9f8382057b62d8dc7d7d514bd1ea1e218f8123

KVGarg:
If I stack 9d9f838 onto d643d24ca3dfd0b8b6effaffa271cdd99f471e03 it will change the hash value 9d9f838 to some new hash as it will treat it as a commit of this PR and based on timestamp it will change hash

KVGarg:
@jayvdb Travis CI is taking just 8min to build it but Netlify is taking more than double (17m 24s) :disappointed:

KVGarg:
Its the morning time usually only 5-6 hrs (6am-12pm IST I guess) available to us in which we can get netlify build green :smile:

jayvdb:
@KVGarg , please fix https://github.com/coala/community/pull/246 .  Then rebase https://github.com/coala/community/pull/244 on top of  https://github.com/coala/community/pull/246 .  Just do it please.

KVGarg:
Hi @jayvdb  can you please tell me what should I fix in #246? Its' already synced with latest moban changes

KVGarg:
Also this PR (#244) is already rebased on top of #246
image

jayvdb:
ya, I see you did fix it, and stacked them correctly.  I was previously responding to https://github.com/coala/community/pull/244#issuecomment-493321481 which sounded like you were doing something else.

Anyways, here goes the merge ...

jayvdb:
Nope.  https://github.com/coala/community/pull/244/commits/0ffb8f599d9639876be5fd496801c5346c7e8b9d commit message doesnt describe the contents of the commit, and it should not be fixes, because the code is not erroneous -- the code is fine, it is merely that Django 2.2 is released and is not supported yet.  In short, it is best to not use `Fixes ` unless the issue has been labelled a bug.

KVGarg:
@jayvdb just noticed that our twitter page on community website redirects to wrong webpage. If we open the preview of a PR (for example - opening the deploy preview of this PR) and then redirect to twitter page on community, the link mentioned on webpage redirects the user to https://twitter.com/coala instead of https://twitter.com/coala_io

jayvdb:
@KVGarg that is not related to this PR.  Create an issue if you have found a bug.

Issue #232, opened 2019-02-04
Deploy of fork to netlify fails the build


KVGarg:
Following the [instructions](https://github.com/coala/community/#deploy-netlify-on-your-fork-) to deploy the changes pushed in forked repo on netlify is failing the build.
Error log - https://gist.github.com/KVGarg/db41b5f3019513f3659dd44c014ba27c

AkshJain99:
@KVGarg  ithink this issue is already opened by  sks444 #189 , isn't it?

KVGarg:
I don't think, it is related to that issue. Because here the netlify build is failing due to non-availability of correct Repository URL.
Check https://gist.github.com/KVGarg/db41b5f3019513f3659dd44c014ba27c#file-community-netlify-build-L226

AkshJain99:
okay but incoorect repository url in the sense what ??  what you think what can be the solution for this?

KVGarg:
> what you think what can be the solution for this?

IDK, I have to debug the code to get the right solution.
Also, I think we need to provide same configuration (environment variables) that are being set while deploying new site on netlify that are already set for master branch on netlify.
Maybe we are missing some information in README.md

KVGarg:
Maybe related to #171

KVGarg:
While deploying a branch on netlify, following problems have been coming listed below being raised from `get_irep()` [function](https://github.com/KVGarg/community/blob/deploy_branch/community/git.py#L111)
1. [hoster](https://github.com/KVGarg/community/blob/deploy_branch/community/git.py#L115) returned as None
2. `Owner` and `Name` attributes in [url](https://github.com/KVGarg/community/blob/deploy_branch/community/git.py#L113) are None due to which error is raised from function [get_repo_slug()](https://github.com/KVGarg/community/blob/deploy_branch/community/git.py#L116)
3. [repo](https://github.com/KVGarg/community/blob/deploy_branch/community/git.py#L117) not an instance of GitHubRepository

KVGarg:
After spending 3-4 hrs on it, finally solved that above solution :open_mouth: 
But as usual failing with building static website :disappointed:

KVGarg:
@jayvdb Can I create a PR for this issue, if it is a valid issue?

KVGarg:
**The local deployed branch on netlify of forked version is passing the build.** So, I guess the issue is no more valid. Maybe to due to node version up-gradation or some packages got upgraded the build is getting passed :+1+
[...]

PR #245, opened 2019-05-10
git.py: Make netlify build pass for local branch

KVGarg:
When the forked community repo is deployed on netlify
the build is used to fail due to non-availability of
attributes name, owner and resource in url attribute.
They're used to have None value instead of a valid
value. So, to resolve this issue I've added a way to
get valid values of these attrs from href attribute in
url object.

Fixes https://github.com/coala/community/issues/232


sks444:
Add a commit message about why it was failing and how did you fixed it.

KVGarg:
The issue is automatically solved for now! Comment https://github.com/coala/community/issues/232#issuecomment-493136867
Closing it as issue is no more valid for now :innocent:

Issue #55, opened 2017-12-18
Create project specific logo


jayvdb:
https://github.com/coala/community/blob/master/docs/_static/images/header.png needs to be removed.

This image was added as a temporary solution for https://github.com/coala/community/issues/10 , as it was a newcomer issue.

This project is a re-usable component that can help any community, so the README should not focus on coala.

It should be replaced with a project specific logo which symbolises creating/supporting/growing communities.

k4ni5h:
can i work on this issue

KVGarg:
> project-specific logo which symbolizes creating/supporting/growing communities.

Thought of this possible idea
![New Logo](https://user-images.githubusercontent.com/35761292/55192745-34f20d00-51cb-11e9-93cc-d56112f26d9c.png)

KVGarg:
Some coala community logo designs https://drive.google.com/drive/folders/113UVZimTZo3rgZqdhq-0a0druyZfQJJH?usp=sharing

PR #249, opened 2019-05-29
README.md: Update header image

KVGarg:
The new header image is designed on the thought
that this project is a re-usable project which
can be forked by any other organisation. Therefore,
the header should be a project specific logo which
symbolises three components - creating, growing and
supporting community.

Closes https://github.com/coala/community/issues/55


sks444:
Hey @li-boxuan, can this be merged?

KVGarg:
> IMO this logo is still a bit coala-specific, as there are many logos related to angular, linux, etc

This is true that I have added programming languages based on languages being used by coala community, but the devs outside circle in the image shows the all the developers having knowledge about all the programming language. If needed, we can have the logos of most common programming languages used today.

> Another concern is whether these logos violate their copyright.

Used  images have no copyright issue, downloaded from freepik.com and after downloading them I've modified them too, to make it completely different from the original ones. the were treated as base images on which the oepn-source community related ideas can be added! So, These images don't violate any kind of copyright.

.travis.yml
```diff
@@ -43,7 +50,7 @@ script:
   - ./.ci/build.sh
   - rm -rf private/
   - pytest
-  - coala --non-interactive -V
+  - coala --flush-cache --non-interactive -V
```
abhishalya, line 53: is this really required?

KVGarg: I guess so, yes! Otherwise the build was failing unnecessarily if the files weren't changed

KVGarg: Even if there are minor changes, like 2-3 lines of code still it was failing. That's why I added this up to re-run the coala on entire project irrespective of cache

PR #254, opened 2019-06-15
 meta-review/: Redesign the webpage

KVGarg:
The redesigned web-page provides a nice
UI/UX design to the web-page with "search"
feature.

Closes https://gitlab.com/coala/GSoC/gsoc-2019/issues/143


sks444:
Do not close GitLab issues from here as mentioned earlier.

templates/meta_review.html
```diff
@@ -1,53 +1,145 @@
...
+    <div class="contributor-image meta-reviewer-image">
+      <img src="//github.com/{{ contributor.login }}.png/?size=210"
+           alt="user-image">
+      <div class="top-right large-font bold-text">
+        {{ contributor.rank }}
+      </div>
+    </div>
+    <div class="metareviewer-details">
+      <div class="contributor-details-part-1 column-flex">
+        <a class="bold-text large-font" href="//github.com
```
li-boxuan: Link is not working for me

KVGarg: I will check if it is feasible to make it case insensitive!
And will also check why the link the isn't working as it is an anchor tag

KVGarg: Made search case insensitive and the link too now is working :+1:

PR #298, opened 2021-05-13
requirements.txt: Use coala IGitt remote as dependency

KVGarg:
The gitmate IGitt repository was bit outdated, and was
using older version of GitHub api requests management.
Since, GitHub updated there api endpoints by depreciating
access token in query params, and recommended to pass
query params in request headers. So, to avoid any future
breaks related to GitHub api, updated coala Gitmate IGitt
repo with updated github token api headers.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ git-url-parse
 django>2.1,<2.2
 django-distill
 django-eventtools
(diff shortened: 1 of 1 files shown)


KVGarg:
@abhishalya Updated requirements file to use coala gitmate repository

abhishalya (approved):

